PR #1, opened 2024-12-04
Use new ResourceType: c_kobo_server

IamJeffG:
Adds new resource type: `c_kobo_server`.  This way we are not pasting API keys everywhere.

The `c_` prefix is technically optional but intended to not clash with resource types from Windmill Hub.

I cannot figure out how to get this type to sync from the wmill CLI.
* Meantime, you can, in the Windmill UI, manually add a new resource type and paste the JSON.
* I suspect the "correct answer" is to [get this type accepted into the Windmill Hub](https://www.windmill.dev/docs/core_concepts/resources_and_types#share-resource-type-on-hub) (or add our own Hub)

## Example
image


f/frizzle/kobo_responses.script.yaml
```diff
@@ -34,18 +33,14 @@ schema:
       description: ''
       default: null
       originalType: string
-    kobo_api_key:
-      type: string
+    kobo_server:
+      type: object
       description: ''
```
rudokemper, line 38: Let's add a description to guide the user about what this resource represents.

f/frizzle/kobo_responses.script.yaml
```diff
@@ -34,18 +33,14 @@ schema:
...
       default: null
       originalType: string
-    kobo_api_key:
-      type: string
+    kobo_server:
```
rudokemper, line 36: I think `kobo_server` could be confusing for cases where a user has multiple accounts (and therefore API keys) on the same server.

I'm thinking of something like `kobo_configuration`. Perhaps with the description of what that actually entails -- e.g. a server/key pairing -- that could be sufficient?

rudokemper (changes requested):

f/frizzle/kobo_responses.script.yaml
IamJeffG: `kobo_account`?  As I look at the Kobo server UI, there is exactly 1 API key per user account.

The other possibility is just `kobo`.  Most of the resource types in the Windmill hub do not specify beyond the service name, even though they actually represent one account or one API Key on that service.

I suggest we go with `kobo_account` because it's less ambiguous, but happy to change.

rudokemper: I'm good with `kobo_account` and a description that describes that this resource includes both a URL and an API key.

IamJeffG:
I've resolved your comments.

Oddly, in this JSON file there is no place to describe the resource type. I have added a description on the form where it's used, though.

rudokemper (approved):

IamJeffG:
I was able to submit this new resource type upstream to the global Windmill Hub: https://hub.windmill.dev/resource_types/193/kobotoolbox_account

Let's see if it syncs down to our deployments!

rudokemper:
Great! For posterity sake, how did you do that?

IamJeffG:
- Instructions: https://www.windmill.dev/docs/misc/share_on_hub#resource-types
- Direct link: https://hub.windmill.dev/resource_types/add

The instructions say my contribution would need to be approved, but I did not see any approval step or delay.

Issue #2, opened 2024-12-04
Port alerts job to run on windmill


IamJeffG:
Port our earth sensing alerts job to run as a windmill script.

Gotchas will include:
- [x] Google credentials file
- [ ] Initial DB table setup

IamJeffG:
> Initial DB table setup

this worked out of the box!

PR #5, opened 2024-12-04
alerts_gcs script accepts GCP credential from a Windmill resource

IamJeffG:
With this PR, the alerts_gcs script accepts Google Cloud service account credential from a Windmill resource.

The Windmill resource type is `gcp_service_account`.

### DIFF
--- a/f/frizzle/alerts_gcs.py
+++ b/f/frizzle/alerts_gcs.py
@@ -132,8 +132,7 @@ def sync_gcs_to_local(
         the (local) file names that were written.
 
     FIXME: Ensure dst_path exists before downloading files.
-    FIXME: sync, don't brute-force re-download files that already exist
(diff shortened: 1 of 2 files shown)


IamJeffG:
Toward #2

f/frizzle/alerts_gcs.py
```diff
@@ -125,7 +135,10 @@ def sync_gcs_to_local(dst_path, bucket_name, territory_id, alerts_metadata_filen
     FIXME: sync, don't brute-force re-download files that already exist
        - see https://github.com/ConservationMetrics/frizzle-sandbox/issues/169
```
rudokemper, line 136: We can move this issue over to here before merging.

rudokemper: https://github.com/ConservationMetrics/gc-scripts-hub/issues/6

rudokemper (approved):

PR #8, opened 2024-12-04
Add `CONTRIBUTING.md` and additional Windmill info for devs

rudokemper:
## Goal

To add `CONTRIBUTING.md` from frizzle-sandbox but modified to match the code needs of this repository, and a few additional notes to the "Development" section of the README.


CONTRIBUTING.md
```diff
@@ -0,0 +1,123 @@
+Thank you for contributing to the GuardianConnector Scripts Hub project!
+Here are some conventions we use to smooth things along.
+
+# Using GitHub
+
+We use GitHub to host code, track issues, and accept pull requests.
+
+## Git Branching
+
+**We build and release from the `main` branch, so code merged here should always be stable.**
```
IamJeffG, line 10:
```suggestion
**We release from the `main` branch, so code merged here should always be stable.**
```

(there is no build)

CONTRIBUTING.md
```diff
@@ -0,0 +1,123 @@
...
+
+> Follow [ESLint Getting Started](https://eslint.org/docs/user-guide/getting-started).
+> To configure ESLint in VS Code, create a `.vscode/settings.json` in the project's root and add
+> ```
+>    "eslint.nodePath": "/home/\${user}/dev/nosh-by-gosh/client/node_modules",
+>    "eslint.workingDirectories": ["client"]
+> ```
+
+
+## Ruff for Python
```
IamJeffG, line 92: Windmill's in-app code editor also does its own linting and ruff auto-formatting.  I have not yet read up on much more than that.

But it occurs to me that whatever auto-formatting Windmill's editor does will need to be matched exactly with local-dev formatting.  For example I am pretty sure that Windmill sets its ruff max line-length to 88 (maybe this is the default, again I haven't looked).

Any ideas how to easily make these sync up?  Some ideas from my end:
- Are windmill's default lint/formatter settings in their version control? We could copy those files.
- Can we set some file in our repo that overrides Windmill's settings?
- Does the `wmill` CLI have anything that formats local code on the windmill server, according to their own rules?


It would be good for one of us to read up on whatever docs or old GH issues Windmill has around this.

README.md
```diff
@@ -22,6 +22,14 @@ because otherwise Windmill tries to make the tests (as with ALL python files) in
 
 ### Development
 
+In Windmill, scripts and flows can be written in Python, TypeScript, Go, and a number of other languages.
+
+The `f/` directory is designated for storing code in a workspace, and will be used when synchronizing the contents of this repository with a server.
```
IamJeffG, line 27: I think what you have is fine. Just a note that (as of what's on `main` now) `wmill.yaml` will also sync a `u/` directory if it exists. This is meant for user-space scripts.  Maybe that doesn't make sense in this repo.

rudokemper: I added a note about this.

CONTRIBUTING.md
```diff
@@ -0,0 +1,123 @@
...
+>      ```
+>      "[python]": {
+>          "editor.formatOnSave": true,
+>          "editor.codeActionsOnSave": {
+>              "source.fixAll": "explicit",
+>              "source.organizeImports": "explicit"
+>          },
+>          "editor.defaultFormatter": "charliermarsh.ruff"
+>      }
+>      ```
```
IamJeffG, line 123:
```suggestion
>      ```

Document Python code according to **[NumPy's documentation
standard](https://numpydoc.readthedocs.io/en/latest/format.html#docstring-standard)**.
```

IamJeffG: unless you have reason to believe that Ruff also  can lint or autoformat docstrings -- in which case we should use that!

rudokemper: I have no particular knowledge about Ruff beyond basic code editor usage and what's been introduced in `frizzle-sandbox`, so no such reason on my part :)

CONTRIBUTING.md
rudokemper: [In the codebase](https://github.com/search?q=repo%3Awindmill-labs%2Fwindmill%20ruff&type=code) it just looks like they run `pip3 install ruff-lsp` in the Dockerfile and provide it as a service through a WebSocket endpoint. 

I asked Windmill's AI service on their Discord:

----

Answer
Based on the available information sources, while Windmill uses Ruff, Black, and Pyright for Python (1)  (2) , the specific linting and formatting settings for these tools are not documented in the provided sources. 

From what we can see in one debug log, Ruff is run with these flags (3) :
```
['--force-exclude', '--no-cache', '--no-fix', '--quiet', '--output-format', 'json', '-', '--stdin-filename']
```

However, I cannot make any claims about other configuration settings as they are not explicitly documented in the information sources. If you need to know about specific linting rules or formatting configurations, I'd suggest clicking the "Ask for help" button to have a team member provide more details.

Sources
[[1, 2] Code assistants](https://www.windmill.dev/docs/code_editor/assistants)
[[3] can not run any python script](https://github.com/windmill-labs/windmill/discussions/3320)

IamJeffG: Wow, thank you so much for doing this digging.

Upon further fiddling with the Windmill in-browser code editor ... it auto-formats Javascript but not Python!  It does provide lint and typehints, but no auto-formatting.

And as you found it doesn't seem that windmill-lsp overrides any defaults.

So -- I believe that what CONTRIBUTING.md says is correct and will not conflict with anything!

README.md
IamJeffG: I agree with what you wrote.  I think to actually not use the `u/` convention, as you've added, you also need to remove `'u/**'`  from `wmill.yaml`'s `includes`:

    includes: ['f/**', 'u/**']

IamJeffG (approved):
pls update wmill.yaml before merge

Issue #9, opened 2024-12-04
[frizzle] Move over / write new e2e tests for scripts


rudokemper:
## Feature Request

The e2e tests in `frizzle_sandbox` are baked into Dagster functionality. Let's figure out how to refactor them for usage here. We can start with one (I'd recommend Kobo) and tackle the others from there.

- [x] Kobo
- [x] CoMapeo
- [x] Alerts

PR #11, opened 2024-12-05
Improve processing logic for CoMapeo script

rudokemper:
## Goal

To improve the processing logic for the CoMapeo Script in several ways (see next section).

## What I changed

* Runs no longer fail if no projects were returned by a successful request to the CoMapeo API. Instead, the remainder of the script is skipped.
* Move the code to rewrite the project name and columns from the `CoMapeoDBWriter` methods to `process_comapeo_data`.
* Finish the work started in `frizzle-sandbox PR` [Implement retry and skip logic for downloading attachments in CoMapeo op graph](https://github.com/ConservationMetrics/frizzle-sandbox/pull/179) by:
  * Implementing the Python `retry` library
  * Tracking attachment failure with an `attachment_failed` var and raising a RuntimeError if true at the end of the script.

## What I'm not doing here

Tests. I would like to do this in a follow-on PR and after we decide on an optimal method for e2e.

## Screenshots

_Relating to the retry logic for attachment download failure_

This incident...

```
INFO:f.frizzle.comapeo.comapeo_observations:[URL REDACTED]
INFO:f.frizzle.comapeo.comapeo_observations:Download completed.
INFO:f.frizzle.comapeo.comapeo_observations:[URL REDACTED]
ERROR:f.frizzle.comapeo.comapeo_observations:Exception during download: Response ended prematurely
ERROR:f.frizzle.comapeo.comapeo_observations:Attachment download failed for URL: [URL REDACTED]. Skipping attachment.
INFO:f.frizzle.comapeo.comapeo_observations:[URL REDACTED]
INFO:f.frizzle.comapeo.comapeo_observations:Download completed.
INFO:f.frizzle.comapeo.comapeo_observations:[URL REDACTED]
INFO:f.frizzle.comapeo.comapeo_observations:Download completed.
```

...will set `attachment_failed` = True, and after the run finishes its processing...

```
INFO:f.frizzle.comapeo.comapeo_observations:Project 2 (ID: rqz5ow1rck6fsty5yh34gxn4h3pjox5mmujctoamq86knu8iuary, name: COP16): Processed 129 observation(s).
INFO:f.frizzle.comapeo.comapeo_observations:Downloaded 179 observations from 2 projects.
INFO:f.frizzle.comapeo.comapeo_observations:Table comapeo_cop16_demo_project created.
INFO:f.frizzle.comapeo.comapeo_observations:Inserting 50 submissions into DB.
INFO:f.frizzle.comapeo.comapeo_observations:Inserting 129 submissions into DB.
INFO:root:Wrote response content to database table(s) with prefix [comapeo]
```

...the script will ultimately throw a RuntimeError causing the run to be assigned a Failed status:

image

## Notes

* With this refactor to remove processing logic from `CoMapeoDBWriter`, this script class is now quite similar to `KoboDBWriter`. The main difference is that, when I wrote this code for `Dagster`, I chose not to include methods related to mappings and the use of a reverse_separator. In the future, however, we could potentially have a single `DBWriter` that serves multiple scripts in a flow.
* In implementing retry logic, I read @IamJeffG's cautionary notes on the use of the `retrying` package...

  > if you use it I strongly recommend reading the docs carefully to make sure it's retrying the bare minimum of what needs to be and only if there is a reasonable chance that retry will succeed.
  >
  > For a sense of what could go "wrong" (and this applies to your current inline implementation too) check https://github.com/ConservationMetrics/frizzle/issues/90

  ...and reviewed the [`retrying`](https://pypi.org/project/retrying/) docs a bit on setting up an optimal configuration for retries. 

  I _have_ anecdotally noticed that attachments downloads from a CoMapeo API still sometimes randomly fail, so I think there is value in the use of retry logic. 

  I opted for a `stop_max_attempt_number`=3 and `wait_exponential_multiplier=500` to set wait times of 500ms, 1000ms, and 2000ms for the first, second, and third attempts, respectively. I'm not convinced these are the right intervals and am happy to receive feedback.


f/frizzle/comapeo/comapeo_observations.py
```diff
@@ -1,16 +1,18 @@
...
 
 import logging
+import mimetypes
 import os
 import re
 
 import psycopg2
-import requests
 from psycopg2 import errors
+import requests
```
IamJeffG, line 13: Was moving this import done automatically and if so by what?

The setup described in https://github.com/ConservationMetrics/gc-scripts-hub/pull/8 (specifically `"source.organizeImports": "explicit"`) would have kept this as it previously was, not made this change.  If Windmill's UI changed this then that's an interesting clash.

f/frizzle/comapeo/comapeo_observations.py
```diff
@@ -37,24 +39,40 @@ def main(
...
-        comapeo_server_base_url, comapeo_access_token, comapeo_projects, attachment_root
+    attachment_failed = False
+
+    comapeo_data, attachment_failed = process_comapeo_data(
+        comapeo_server_base_url,
+        comapeo_access_token,
+        comapeo_projects,
+        db_table_prefix,
+        attachment_root,
+        attachment_failed,
```
IamJeffG, line 60: I see no reason that `attachment_failed` has to be sent from outside.  It returns the boolean and that should be enough.

f/frizzle/comapeo/comapeo_observations.py
```diff
@@ -65,17 +83,15 @@ def fetch_comapeo_projects(
     logger.info("Fetching projects from CoMapeo API...")
     response = requests.request("GET", url, headers=headers, data=payload)
 
-    if response.status_code == 404:
-        logger.error(f"URL not found: {url}")
+    if response.status_code != 200:
+        logger.error(f"Unexpected status code: {response.status_code}")
         raise ValueError(
             f"Failed to fetch projects: {response.status_code} - {response.reason}"
         )
```
IamJeffG, line 90:
```suggestion
    response.raise_for_status()
```


[FYI this method exists.](https://3.python-requests.org/user/quickstart/#response-status-codes)

Don't accept the suggestion if you think there's good value in this extra formatting of the error that you add.

f/frizzle/comapeo/comapeo_observations.py
```diff
@@ -282,26 +316,13 @@ def handle_output(self, outputs):
         """
```
IamJeffG, line 316: so this method now writes to multiple tables?

would be good to write that in the docstring.

f/frizzle/comapeo/comapeo_observations.py
```diff
@@ -92,20 +108,50 @@ def fetch_comapeo_projects(
...
-    if not comapeo_projects:
-        raise ValueError("No valid projects found.")
-
     return comapeo_projects
 
 
+@retry(
+    stop_max_attempt_number=3,
+    wait_exponential_multiplier=500,
+)
```
IamJeffG, line 117: I forget - if this fails 3 times in the row, what happens?  Will the entire script fail?  Or does it silently move on?

f/frizzle/comapeo/comapeo_observations.py
```diff
@@ -92,20 +108,50 @@ def fetch_comapeo_projects(
...
+    try:
+        response = requests.get(url, headers=headers)
+
+        content_type = response.headers.get("Content-Type", "")
+        extension = mimetypes.guess_extension(content_type) or ""
+
+        file_name = os.path.basename(url) + extension
+        os.makedirs(os.path.dirname(save_path), exist_ok=True)
+        with open(save_path, "wb") as file:
+            file.write(response.content)
```
IamJeffG, line 128:
```suggestion
        with open(save_path, "wb") as f:
            f.write(response.content)
```

`file` is a reserved word in Python

f/frizzle/comapeo/comapeo_observations.py
```diff
@@ -92,20 +108,50 @@ def fetch_comapeo_projects(
...
     return comapeo_projects
 
 
+@retry(
+    stop_max_attempt_number=3,
+    wait_exponential_multiplier=500,
+)
+def download_attachment(url, headers, save_path):
+    try:
+        response = requests.get(url, headers=headers)
```
IamJeffG, line 120:
```suggestion
        response = requests.get(url, headers=headers)
        response.raise_for_status()
```

f/frizzle/comapeo/comapeo_observations.py
```diff
@@ -92,20 +108,50 @@ def fetch_comapeo_projects(
...
+        logger.info("Download completed.")
+        return file_name
+
+    except Exception as e:
+        logger.error(f"Exception during download: {e}")
+        return None
+
+
+def camel_to_snake(name):
+    return re.sub("([a-z0-9])([A-Z])", r"\1_\2", name).lower()
```
IamJeffG, line 138: To get into the practice I suggest adding a quick unit test for this function.

1. `comapeo_observations_test.py` naming convention will be ignored by `wmill sync`
2. modify `tox.ini` to have a new section `[testenv:comapeo_observations]`

Feel free to do this in another PR; or we just agree to handle it in #9

f/frizzle/comapeo/comapeo_observations.py
```diff
@@ -92,20 +108,50 @@ def fetch_comapeo_projects(
...
+
 def process_comapeo_data(
-    comapeo_server_base_url, comapeo_access_token, comapeo_projects, attachment_root
+    comapeo_server_base_url,
+    comapeo_access_token,
+    comapeo_projects,
+    db_table_prefix,
+    attachment_root,
+    attachment_failed,
 ):
```
IamJeffG, line 148: please add a docstring, see for example the one I added at https://github.com/ConservationMetrics/gc-scripts-hub/blob/9a134d0315c3bb07f69f87e9cd9a8d8e1984e873/f/frizzle/alerts/alerts_gcs.py#L214-L243

More generally, I'd say that a verb like "process" is almost devoid of useful meaning in computer science (all we do is process data, for various definitions of "process" and various examples of "data").  Can this be made more precise?

If not, is it a sign that this function is doing too much?  (NB: reviewing this code, I can also say this function is likely doing too much)

Food for thought; don't sweat all that in this PR.  Just do add the docstring.

IamJeffG: It seems I failed to mention the preferred docstring format in `CONTRIBUTING.md` ...   I'll comment over on #8.

f/frizzle/comapeo/comapeo_observations.py
```diff
@@ -92,20 +108,50 @@ def fetch_comapeo_projects(
...
     comapeo_data = {}
     for index, project in enumerate(comapeo_projects):
         project_id = project["project_id"]
         project_name = project["project_name"]
         sanitized_project_name = re.sub(r"\W+", "_", project_name).lower()
+        prefixed_project_name = f"{db_table_prefix}_{sanitized_project_name}"
```
IamJeffG, line 154: Maybe can wait for later, but would we ever allow the `db_table_prefix` to be empty? In which case we should not prepend the `_`.

f/frizzle/comapeo/comapeo_observations.py
rudokemper: No, I moved it because I thought it made more sense to follow `psycopg2` related imports.

f/frizzle/comapeo/comapeo_observations.py
rudokemper: No. What happens is what I was describing in "Screenshots" in the PR body.

The script moves on after 3 failed retries, but tracks the failure by setting `attachment_failed` = true. The script then throws a RuntimeError at the end if that's the case, which therefore marks the run as Failed.

f/frizzle/comapeo/comapeo_observations.py
rudokemper: As I wrote in the PR body, I wanted to handle unit tests in a different PR.

f/frizzle/comapeo/comapeo_observations.py
rudokemper: When I first wrote this code for Dagster, I opted to hem closely to the KoboToolbox op graph, which also has a similarly named function `process_forms`, hence this name. But I agree with you.

I'm going to rename it to `download_and_transform_comapeo_data` for now, and will add docstrings to each function.

f/frizzle/comapeo/comapeo_observations.py
IamJeffG: Thank you. Ok, that part makes sense to me now -- and so my confusion is actually something different.

Can you confirm: Is it actually retrying?  I'd claim it's not.

I thought `@retry` would retry if it catches an exception, but you never give it the opportunity to.  Line 132 catches & silences any exception that might have come from this function.

f/frizzle/comapeo/comapeo_observations.py
IamJeffG: Apologies :-)

f/frizzle/comapeo/comapeo_observations.py
rudokemper: You're right, it's not. It seemed like it did when I ran it with some attachments I knew would fail, but I just confirmed with logging that it only runs once.

In that case, I'm doubting whether `@retry` is the solution here. Because if I don't catch & silence the exception, it will cause the script to fail (and not proceed) right away after 3 attempts.

...I'm also increasingly sympathetic to your line of reasoning [here](https://github.com/ConservationMetrics/frizzle/issues/90):

> This is ridiculous that we even retry. All this code is polling code, so even if it fails it will retry again after the poll interval. I'm going to just rip out the retries instead of dealing with all the desired behavior up above.

I'm tempted to drop the retry and just keep the `attachment_failed` bool.

rudokemper:
Apologies, I did not commit each change per suggestion, but all feedback except retry logic has been addressed (thank you!)

f/frizzle/comapeo/comapeo_observations.py
IamJeffG: I'd be ok with that.

IamJeffG (approved):

PR #13, opened 2024-12-05
Tests now get their own directories

IamJeffG:
## Goal

The previous convention for organizing tests — tests must end in `_tests.py` — was too restrictive.  This PR allows source control to contain any sort of test assets, such as `conftest.py`, images, JSON files.  And also tests for other programming languages.
 

## What I changed
It achieves this goal by establishing convention (and `wmill sync` exclude rule) where any folder called `./tests/` is ignored from sync.

This new pattern is illustrated by inserting a supporting file `requirements-test.txt` at `f/frizzle/kobo/tests/requirements-test.txt`. This would not have been possible before.


rudokemper (approved):
Excellent!

PR #15, opened 2024-12-05
Streamline, and add end-to-end test for, kobo_responses

IamJeffG:
## Goal

end-to-end test for `kobo_responses` script

Toward #9 

## What I changed

1. [kobo_responses no longer fetches ALL forms on server](https://github.com/ConservationMetrics/gc-scripts-hub/commit/7be7475f9e0bd707c57f6147f5352b5ec7404410). I'm open to making this a separate PR, but i found this helpful so I could mock less data for the test.
2. end-to-end test
     - add  test assets: mocked responses from a kobotoolbox server
     - use `responses` python package to mock those requests, in a pytest fixture

The end-to-end test sets up a little postgresql server for testing.  This is somewhat heavyweight, and also requires people running the test (including the CircleCI worker) to have `postgresql` installed.  I've added a code comment that we could, perhaps, eliminate this requirement, by converting our scripts to use SqlAlchemy so that, while PostgreSQL is used in production, tests could use SQLite.

## Notes for reviewers

Turn "hide whitespace" on :-)


f/frizzle/kobo/kobo_responses.py
```diff
@@ -34,92 +34,72 @@ def main(
...
+                        with open(save_path, "wb") as file:
+                            file.write(response.content)
+                        logger.debug(
+                            "Download completed: " + attachment["download_url"]
+                        )
+                    else:
+                        # TODO: add retries
+                        logger.error("Failed downloading attachments.")
+
+    logger.info(
```
rudokemper, line 99: Currently, we differ on where we log success messages for the different steps in the data flow. In comapeo (following suit from the previous convention in kobo), [they are in `main`](https://github.com/ConservationMetrics/gc-scripts-hub/blob/d38924208895f4a7536ecfaf807eadb3d0ff3c23/f/frizzle/comapeo/comapeo_observations.py#L40-L56). Here I see you've opted to place the log in the function itself, which I agree is the better choice. We could consider doing the same for the DBWriter.

Anyway, I know that's not the point here. We can address this (and other kobo script modifications) in another PR.

rudokemper (approved):
Nevermind my comment; we can clean up the Kobo script itself, and get aligned on common patterns across the scripts, in a separate PR. Tests look great, thanks!

PR #16, opened 2024-12-06
Add end-to-end test for alerts_gcs

IamJeffG:
## Goal

end-to-end test for `alerts_gcs` script

Toward #9 

## What I changed

1. The test uses the a GCS emulator that runs in a Docker container. Thus I:
    - install `tox-docker` and configure the emulator in `tox.ini`
    - refactor to create a `_main()` method that allows injecting an already instantiated GCS Client.
2. New pytest.fixture `mock_alerts_storage_client` that pre-populates the emulated storage with files (binary and text) of mock alerts.  These files are stored in `f/frizzle/alerts/tests/assets/`

As with all frizzle data flows' tests, I've added it as a new tox.ini "environment"


IamJeffG:
I will need to take a deeper look why CircleCI is failing this.  Tests pass locally, but I could imagine the new tox-docker addition is not what CI is expecting.

rudokemper (approved):
Failing tests notwithstanding, LGTM.

PR #18, opened 2024-12-09
sharing fixtures across tests

IamJeffG:
## Goal

Allow sharing pytest fixtures across tests suites, even when each suite runs in different install environments.

## What I changed

Many Frizzle flows follow similar patterns, and therefore use the same pytest fixtures in their test suites.  This PR builds a library of "sharable" fixtures that all suites can avail themselves.  A `*_test.py` file can make this library available by including the line

    pytest_plugins = ["f.frizzle.tests.conftest"]

See https://docs.pytest.org/en/stable/how-to/writing_plugins.html

## Notes for reviewers

This PR may be ill-advised, for up to a few reasons:

0. This is my first use of "pytest plugins" so not sure if I have the mental model correct.
1. It introduces some indirection that I don't love.  `gc-scripts-hub` is already unconventional in that `tox.ini` builds separate environments different test suites, since their dependencies can diverge so much.  Under this PR, one test suite may use the plugin that introduces its own package dependencies but those dependencies are not declared at the shared library (plugin) scope.  Rather the test suite needs to be aware of which shared fixture(s) it uses -- and that fixture's package dependencies -- and declare them in its own `requirements-test.txt`.
2. Probably long term I will have wanted to define these shared fixtures at even a lower scope: e.g. one file per fixture.  And then a specific test file would do something more along the lines of

```
pytest_plugins = ["f.frizzle.tests.plugins.postgresql", "f.frizzle.tests.plugins.mocked_responses"]
```

### DIFF
--- a/f/frizzle/tests/conftest.py
+++ b/f/frizzle/tests/conftest.py
@@ -1,0 +1,20 @@
+import pytest
+import responses
+import testing.postgresql
+
+
+@pytest.fixture
+def mocked_responses():
+    """responses.RequestsMock context, for testing code that makes HTTP requests."""
+    with responses.RequestsMock() as rsps:
+        yield rsps
+
+
+@pytest.fixture
+def pg_database():
+    """A dsn that may be used to connect to a live (local for test) postgresql server"""
+    db = testing.postgresql.Postgresql(port=7654)
+    dsn = db.dsn()
+    dsn["dbname"] = dsn.pop("database")
+    yield dsn
+    db.stop
--- a/f/frizzle/tests/conftest.py
+++ b/f/frizzle/tests/conftest.py
@@ -1,11 +1,11 @@
 import pytest
-import responses
-import testing.postgresql
 
 
 @pytest.fixture
 def mocked_responses():
     """responses.RequestsMock context, for testing code that makes HTTP requests."""
+    import responses
+
     with responses.RequestsMock() as rsps:
         yield rsps
 
@@ -13,6 +13,8 @@ def mocked_responses():
 @pytest.fixture
 def pg_database():
     """A dsn that may be used to connect to a live (local for test) postgresql server"""
+    import testing.postgresql
+
     db = testing.postgresql.Postgresql(port=7654)
     dsn = db.dsn()
     dsn["dbname"] = dsn.pop("database")
--- a/f/frizzle/kobo/tests/conftest.py
+++ b/f/frizzle/kobo/tests/conftest.py
@@ -2,16 +2,10 @@ import re
 from dataclasses import dataclass
 
 import pytest
-import responses
-import testing.postgresql
 
 from f.frizzle.kobo.tests.assets import server_responses
 
-
-@pytest.fixture
-def mocked_responses():
-    with responses.RequestsMock() as rsps:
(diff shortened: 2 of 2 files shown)


IamJeffG:
Throwing this idea out into the world for discussion.  Happy to receive criticism or rejection on it if warranted.

f/frizzle/tests/conftest.py
```diff
@@ -0,0 +1,20 @@
+import pytest
+import responses
```
IamJeffG, line 2: From the PR description:

> It introduces some indirection that I don't love. gc-scripts-hub is already unconventional in that tox.ini builds separate environments different test suites, since their dependencies can diverge so much. Under this PR, one test suite may use the plugin that introduces its own package dependencies but those dependencies are not declared at the shared library (plugin) scope. Rather the test suite needs to be aware of which shared fixture(s) it uses -- and that fixture's package dependencies -- and declare them in its own `requirements-test.txt`.

This line is an example.  Merely pulling in this file will give an `ImportError` unless the test suite itself has opted to install `responses` and `testing.postgresl`.


This could be ameliorated by pulling the `import` into the fixture itself, so that a test only needs to install the fixtures it opts to use. Dunno if that solution is good enough (I'm ok with it for now..) or is just  a warning sign of things to come.

rudokemper:
Interesting proposal. On one hand, it reduces some redundancy in common Frizzle fixtures, but there's something to be said for having everything I need to know about a script be contained within its Frizzle subdirectory. That is to say, if I only care about `f/frizzle/kobo`, then I don't need to look at anything else. 

Maybe one way of thinking about that is if this proposal is poised to raise the barrier of contribution for a new contributor in a slight way. I don't think so if we provide good documentation about why these fixtures (e.g. connect to a postgres db) are being shared. If anything, it's a good opportunity for us to describe what "frizzle" is, and why these scripts are grouped together to begin with. (I've got a forthcoming PR adding a README with some still-relevant copy from past repos).

IamJeffG:
> Interesting proposal. On one hand, it reduces some redundancy in common Frizzle fixtures, but there's something to be said for having everything I need to know about a script be contained within its Frizzle subdirectory. That is to say, if I only care about f/frizzle/kobo, then I don't need to look at anything else.

I think this is a good aim, and this PR goes against it.

The only way I can imagine to do something like this PR in-line with your guidance is to package this up as its own pytest plugin, distributed to a test environment through means other than local path search.  Yet in fact there's not very much that's cohesive about the code that this PR groups together in the first place, so I struggle to justify doing that.

I am going to close this without merge.

The greater question that this PR attempted to address is "how to deal with shared code between scripts?"  This PR only dealt with test code but we do have the same problem in script code, for example see #25

> Maybe one way of thinking about that is if this proposal is poised to raise the barrier of contribution for a new contributor in a slight way. I don't think so if we provide good documentation about why these fixtures (e.g. connect to a postgres db) are being shared. If anything, it's a good opportunity for us to describe what "frizzle" is, and why these scripts are grouped together to begin with. (I've got a forthcoming PR adding a README with some still-relevant copy from past repos).

I agree that the documentation propose will be nice.  I am not convinced that it actually resolves the problems with this PR, though.

PR #19, opened 2024-12-09
Add high-level and script documentation for frizzle

rudokemper:
## Goal

Closes https://github.com/ConservationMetrics/gc-scripts-hub/issues/10.

To pilfer and retool some of the still-relevant documentation from past frizzle attempts, and to add some new copy.

## What I changed

Added a `README.md` for each frizzle script, and a relatively short one for `f/frizzle` that we may expand in the future with information about things like shared fixtures (if we decide to go that route).

In the script READMEs, I kept the old frizzle v1 convention for documenting config. I realize there is some potential redundancy between that and the description strings in the `script.yaml` files, but the difference is nevertheless salient: the latter is what the end user sees on Windmill, and the information in the READMEs is for developers or folks running the scripts locally or elsewhere.


rudokemper:
Noting one or more var name changes for `alerts` in https://github.com/ConservationMetrics/gc-scripts-hub/pull/16, and will ensure all vars documented here correspond to the latest batch of PRs about tests / script improvement.

rudokemper:
@IamJeffG I have adapted this PR in line with what we discussed.

I didn't touch the alerts `script.yaml` because I know you have updated that file in your alerts e2e test PR.

I also didn't add additional docstrings to any scripts. I will document doing so for kobo in a ticket also focused on cleanup.

IamJeffG:
I've pushed a new commit to this branch that adds the input descriptions for `alerts_gcs.py`.

IamJeffG (approved):
I tweaked some prose.   Feel free to ask for justification of why I propose these changes.

Otherwise, looks great, thanks for doing it.

PR #35, opened 2024-12-12
Add `postgres_to_geojson` export script

rudokemper:
_Note: there is absolutely no rush in reviewing this. I felt inspired to try this out this morning, but there is no immediate need to have this merged._

## Goal

The objective here is twofold:

1. To explore and generate discussion about how Windmill can be utilized for unscheduled, non-"frizzle" tasks for GuardianConnector usage.
2. In so doing, to use a real task pulled from our [private `gc-programs`](https://github.com/ConservationMetrics/gc-programs/tree/main/scripts/postgres-to-geojson-converter) repo that I run at least a few times a year for one of our community partners: generating a GeoJSON file from a database table that has geometry columns (that is, not PostGIS `geometry`, but `g__type` and `g__coordinates`).

The vision here is that when a community partner requests to have a spatial data file generated from one of their db tables (e.g. CoMapeo data or KoboToolbox data with geometry columns), a Windmill user can run this script as a one-off to generate a GeoJSON file on their datalake. The GeoJSON export can then be accessed, and shared, via FileBrowser. No more need for running a script in command line (the status quo).

## Screenshots

image

## What I changed

1. Created an `f/export/` dir to store this script and any other export scripts we may want to create. I am not convinced this is the best name and open to suggestions.
2. Added a `postgres_to_geojson` script with all of the expected Windmill bells and whistles. It mostly follows the same operations as the script in `gc-programs`, but I added some error handling and documentation.
3. Added an e2e test for the script.

### DIFF
--- a/f/export/folder.meta.yaml
+++ b/f/export/folder.meta.yaml
@@ -1,0 +1,6 @@
+summary: null
+display_name: export
+extra_perms:
+  u/rudo: true
+owners:
+  - u/rudo
--- a/f/export/folder.meta.yaml
+++ b/f/export/folder.meta.yaml
@@ -1,2 +1,1 @@
-summary: A collection of export scripts
 display_name: export
--- a/f/export/folder.meta.yaml
+++ b/f/export/folder.meta.yaml
@@ -1,6 +1,2 @@
-summary: null
+summary: A collection of export scripts
 display_name: export
-extra_perms:
-  u/rudo: true
-owners:
-  - u/rudo
--- a/f/export/postgres_to_geojson/postgres_to_geojson.py
+++ b/f/export/postgres_to_geojson/postgres_to_geojson.py
@@ -85,7 +85,7 @@ def format_data_as_geojson(data):
         geometry = {}
         feature_id = None
 
-        # The expectated schema here is that geometry columns are prefixed with "g__"
+        # The expected schema here is that geometry columns are prefixed with "g__"
         # If an "_id" column is present, it is used as the feature id
         # All other columns are treated as properties
         for col, value in zip(columns, row):
--- a/f/export/postgres_to_geojson/postgres_to_geojson.py
+++ b/f/export/postgres_to_geojson/postgres_to_geojson.py
@@ -35,6 +35,19 @@ def main(
 
 
 def fetch_data_from_postgres(db_connection_string, table_name: str):
+    """
+    Fetches all data from a specified PostgreSQL table.
+
+    Parameters
+    ----------
+        db_connection_string (str): The connection string for the PostgreSQL database.
+        table_name (str): The name of the table to fetch data from.
+
+    Returns
+    -------
+        tuple: A tuple containing a list of column names and a list of rows fetched from the table.
+    """
+
     try:
         conn = psycopg2.connect(db_connection_string)
         cursor = conn.cursor()
@@ -43,7 +56,7 @@ def fetch_data_from_postgres(db_connection_string, table_name: str):
         rows = cursor.fetchall()
     except psycopg2.Error as e:
         logger.error(f"Error fetching data from {table_name}: {e}")
-        return None, None
+        raise
     finally:
         cursor.close()
         conn.close()
@@ -56,12 +69,15 @@ def format_data_as_geojson(data):
     """
     Converts data from a PostgreSQL table into a GeoJSON FeatureCollection.
 
-    Args:
+    Parameters
+    ----------
         data (tuple): A tuple containing columns and rows fetched from the database.
 
-    Returns:
+    Returns
+    -------
         dict: A GeoJSON FeatureCollection with features extracted from the data.
     """
+
     columns, rows = data
     features = []
     for row in rows:
@@ -99,6 +115,16 @@ def format_data_as_geojson(data):
 
 
 def save_file(data, db_table_name: str, storage_path: str):
+    """
+    Saves the provided data as a GeoJSON file in the specified storage path.
+
+    Parameters
+    ----------
+        data (dict): The data to be saved, formatted as a GeoJSON FeatureCollection.
+        db_table_name (str): The name of the database table, used to name the output file.
+        storage_path (str): The directory path where the GeoJSON file will be saved.
+    """
+
     try:
         os.makedirs(storage_path, exist_ok=True)
         geojson_path = os.path.join(storage_path, f"{db_table_name}.geojson")
--- a/f/export/postgres_to_geojson/postgres_to_geojson.py
+++ b/f/export/postgres_to_geojson/postgres_to_geojson.py
@@ -1,0 +1,110 @@
+# requirements:
+# psycopg2-binary
+
+import json
+import logging
+import os
+
+import psycopg2
+
+# type names that refer to Windmill Resources
+postgresql = dict
+
+logging.basicConfig(level=logging.INFO)
+logger = logging.getLogger(__name__)
+
+
+def conninfo(db: postgresql):
+    """Convert a `postgresql` Windmill Resources to psycopg-style connection string"""
+    # password is optional
+    password_part = f" password={db['password']}" if "password" in db else ""
+    conn = "dbname={dbname} user={user} host={host} port={port}".format(**db)
+    return conn + password_part
+
+
+def main(
+    db: postgresql,
+    db_table_name: str,
+    storage_path: str = "/frizzle-persistent-storage/datalake/export",
+):
+    data = fetch_data_from_postgres(conninfo(db), db_table_name)
+
+    feature_collection = format_data_as_geojson(data)
+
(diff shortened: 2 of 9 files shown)


f/export/folder.meta.yaml
```diff
@@ -0,0 +1,2 @@
+summary: A collection of export scripts
```
rudokemper, line 1: I'm not sure where this shows up. I can't find it back on the Windmill UI, so I might omit.

Also, note the omission of `extra_perms` and `owners`. Syncing works fine without these. So we can remove them from frizzle too (unless there was a good reason to have them).

f/export/postgres_to_geojson/postgres_to_geojson.py
```diff
@@ -0,0 +1,136 @@
...
+
+    Returns
+    -------
+        tuple: A tuple containing a list of column names and a list of rows fetched from the table.
+    """
+
+    try:
+        conn = psycopg2.connect(db_connection_string)
+        cursor = conn.cursor()
+        cursor.execute(f"SELECT * FROM {table_name};")
```
IamJeffG, line 54: please follow the pattern of #37.

```suggestion
        cursor.execute(sql.SQL("SELECT * FROM {table_name}").format(table_name=sql.Identifier(table_name)))
```

f/export/postgres_to_geojson/postgres_to_geojson.py
```diff
@@ -0,0 +1,136 @@
...
+        columns = [desc[0] for desc in cursor.description]
+        rows = cursor.fetchall()
+    except psycopg2.Error as e:
+        logger.error(f"Error fetching data from {table_name}: {e}")
+        raise
+    finally:
+        cursor.close()
+        conn.close()
```
IamJeffG, line 62: At some point in the near future I  am going to go on an advocacy crusade for better practices around connection and cursor management. Namely upgrading to psycopg v3 and using connection  managers.

Let's not bother with that here. I'll open a separate issue and create an example.

f/export/folder.meta.yaml
IamJeffG: I did not have a good reason at the time.

One question: Once you remove them from frizzle.meta.yaml, does that actually end up changing the permissions in a way we don't want?  I'd imagine it could.

f/export/postgres_to_geojson/postgres_to_geojson.py
```diff
@@ -0,0 +1,136 @@
...
+def main(
+    db: postgresql,
+    db_table_name: str,
+    storage_path: str = "/frizzle-persistent-storage/datalake/export",
+):
+    data = fetch_data_from_postgres(conninfo(db), db_table_name)
+
+    feature_collection = format_data_as_geojson(data)
+
+    save_file(feature_collection, db_table_name, storage_path)
```
IamJeffG, line 34: One day the quantity of data will be too much to hold into memory, so we will want to adapt a streaming solution.  No need to touch it now tho.

ALSO - I wonder if this script could be simplified to a thin wrapper over an `ogr2ogr` call.  There are postgres and geojson drivers for ogr2ogr.  I don't know if having just dropped the PostGIS extension makes the postgres driver a non-starter, but we could always add it back.

f/export/postgres_to_geojson/postgres_to_geojson.py
```diff
@@ -0,0 +1,136 @@
...
+        # The expected schema here is that geometry columns are prefixed with "g__"
+        # If an "_id" column is present, it is used as the feature id
+        # All other columns are treated as properties
+        for col, value in zip(columns, row):
+            if col == "_id":
+                feature_id = value
+            elif col.startswith("g__"):
+                geometry[col[3:]] = value
+            else:
+                properties[col] = value
```
IamJeffG, line 97: :+1:

f/export/postgres_to_geojson/postgres_to_geojson.py
```diff
@@ -0,0 +1,136 @@
...
+        storage_path (str): The directory path where the GeoJSON file will be saved.
+    """
+
+    try:
+        os.makedirs(storage_path, exist_ok=True)
+        geojson_path = os.path.join(storage_path, f"{db_table_name}.geojson")
+        with open(geojson_path, "w") as f:
+            json.dump(data, f)
+    except OSError as e:
+        logger.error(f"Error saving file {geojson_path}: {e}")
```
IamJeffG, line 134: Do you really want to swallow this exception?  As-is I think the script will succeed after failing to write the file.

I'd drop the previous 2 lines.  If there's an OSError, the script will fail, which seems more like what we want.

f/export/postgres_to_geojson/postgres_to_geojson.py
```diff
@@ -0,0 +1,136 @@
...
+    ----------
+        data (dict): The data to be saved, formatted as a GeoJSON FeatureCollection.
+        db_table_name (str): The name of the database table, used to name the output file.
+        storage_path (str): The directory path where the GeoJSON file will be saved.
+    """
+
+    try:
+        os.makedirs(storage_path, exist_ok=True)
+        geojson_path = os.path.join(storage_path, f"{db_table_name}.geojson")
+        with open(geojson_path, "w") as f:
```
IamJeffG, line 131: No need to change, but as a learning opportunity I encourage you to look into [`pathlib`](https://docs.python.org/3/library/pathlib.html) for easier path manipulation.

```suggestion
        storage_path =  Path(storage_path)
        storage_path.mkdir(parents=True, exist_ok=True)
        geojson_path = storage_path / f"{db_table_name}.geojson"
        with geojson_path.open("w") as f:
```

f/export/postgres_to_geojson/tests/conftest.py
```diff
@@ -0,0 +1,48 @@
...
+        );
+    """)
+
+    # Insert the provided data
+    cursor.execute("""
+        INSERT INTO comapeo_data VALUES
+        ('doc_id_1', '-33.8688', 'False', '2024-10-14T20:18:14.206Z', '2024-10-14T20:18:14.206Z', 'Forest Expedition', 'water', 'Point', '[151.2093, -33.8688]', 'capybara.jpg', 'Rapid', 'forest_expedition', NULL, '151.2093'),
+        ('doc_id_2', '48.8566', 'False', '2024-10-15T21:19:15.207Z', '2024-10-15T21:19:15.207Z', 'River Mapping', 'animal', 'Point', '[2.3522, 48.8566]', 'capybara.jpg', 'Capybara', 'river_mapping', 'capybara', '2.3522'),
+        ('doc_id_3', 35.6895, 'False', '2024-10-16T22:20:16.208Z', '2024-10-16T22:20:16.208Z', 'Historical Site', 'location', 'Point', '[139.6917, 35.6895]', NULL, 'Former village site', 'historical', NULL, '139.6917');
+    """)
```
IamJeffG, line 41: Nitpick:

I like to deconstruct into 2 separate fixtures:

`pg_database` is just the database, as in our other tests.

Then a second like

```python
@pytest.fixture
def database_mock_data(pg_database):
    conn = psycopg2.connect(**pg_database) 
    cursor.execute(...)
    yield dsn

def my_test_case(database_mock_data):
    # ...
```

So conceptually, the first is just any ol' database; the second is a specific database with the mock data for your test.

f/export/postgres_to_geojson/postgres_to_geojson.py
rudokemper: Interesting. From a cursory read of the documentation https://gdal.org/en/stable/drivers/vector/pg.html it does seem that some ogr2ogr operations are supported without PostGIS installed. 

I propose we keep this in mind for a future reconsideration of how we handle geospatial db data across the stack, writ large.

rudokemper: >One question: Once you remove them from frizzle.meta.yaml, does that actually end up changing the permissions in a way we don't want? I'd imagine it could.

So these lines get added when you run `wmill sync pull`. When I then remove the lines, and run `wmill sync push`, I see no impact on permissions on the server: I'm able to run, schedule, and edit scripts all the same, schedules are still up, and I see no changes in the UI otherwise. Maybe this is different for a regular admin (not superadmin). But for now I don't see a reason to keep these hard-coded.

f/export/folder.meta.yaml
rudokemper: >One question: Once you remove them from frizzle.meta.yaml, does that actually end up changing the permissions in a way we don't want? I'd imagine it could.

So these lines get added when you run `wmill sync pull`. When I then remove the lines, and run `wmill sync push`, I see no impact on permissions on the server: I'm able to run, schedule, and edit scripts all the same, schedules are still up, and I see no changes in the UI otherwise. Maybe this is different for a regular admin (not superadmin). But for now I don't see a reason to keep these hard-coded.

PR #36, opened 2024-12-12
alerts CREATE TABLE with proper SQL data types, drop PostGIS dep

IamJeffG:
## Goal

Drop dep on PostGIS extension. The only thing it was used for was writing a "geom" column, which is never used downstream.

The PostGIS dependency ends up blocking PostgreSQL database version upgrades on Azure hosted pg, because of a version clash.

## What I changed

* Remove "geom" column from alerts table.

While doing that I discovered that all column types (including "geom") were all being set as TEXT, thus the PostGIS dependency itself was completely implicit. Therefore I also had to fix this:
* `CREATE TABLE alerts` and `CREATE TABLE alerts__metadata` now specify all columns up-front, with their correct types.

## What I'm not doing here

There's a possibility of SQL injection in the CREATE TABLE. I will fix that for all scripts in a follow-up PR: #37

### DIFF
--- a/f/frizzle/alerts/alerts_gcs.py
+++ b/f/frizzle/alerts/alerts_gcs.py
@@ -410,7 +410,7 @@ class AlertsDBWriter:
                     _id character varying(36) NOT NULL,
                     -- These are found in "properties" of an alert Feature:
                     alert_type text,
-                    area_alert_ha double precision,
+                    area_alert_ha double precision,  -- only present for polygon
                     basin_id bigint,
                     count bigint,
                     date_end_t0 text,
--- a/f/frizzle/alerts/alerts_gcs.py
+++ b/f/frizzle/alerts/alerts_gcs.py
@@ -100,7 +100,7 @@ def _main(
     outputs = load_alerts_to_postgis(destination_path, gjson_and_metas)
 
     db_writer = AlertsDBWriter(conninfo(db), db_table_name)
-    db_writer.handle_output(outputs)
+    db_writer.handle_output(outputs["geojsons"], outputs["alerts_metadata"])
     logger.info(f"Wrote response content to database table [{db_table_name}]")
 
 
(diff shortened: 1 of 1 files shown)


f/frizzle/alerts/alerts_gcs.py
```diff
@@ -379,106 +379,67 @@ def _table_exists(self, cursor, table_name):
...
                 """
                 cursor.execute(query)
                 conn.commit()
                 logger.info(f"Table {table_name} created.")
-            else:
-                return
-            for column in alert_cols:
-                try:
-                    cursor.execute(f"""
-                    ALTER TABLE {table_name}
```
IamJeffG, line 467: An implication of removing this is that we no longer persist never-before-seen columns that may appear in the alerts.

If/when those show up, we insert the rest of the row but throw away the new field.

f/frizzle/alerts/alerts_gcs.py
```diff
@@ -379,106 +379,67 @@ def _table_exists(self, cursor, table_name):
...
                 query = f"""
-                CREATE TABLE public.{table_name}();
+                CREATE TABLE {table_name}
+                (
+                    _id character varying(36) NOT NULL,
+                    -- These are found in "properties" of an alert Feature:
+                    alert_type text,
+                    area_alert_ha double precision,
+                    basin_id bigint,
+                    count bigint,
```
rudokemper, line 415: Can we add `confidence int`? I was just reviewing the GeoJSON files and seeing this is one that was added by the provider.

f/frizzle/alerts/alerts_gcs.py
```diff
@@ -379,106 +379,67 @@ def _table_exists(self, cursor, table_name):
...
-            if not self._table_exists(cursor, f"{table_name}__metadata"):
+            if not self._table_exists(cursor, table_name):
                 query = f"""
-                CREATE TABLE public.{table_name}();
+                CREATE TABLE {table_name}
+                (
+                    _id character varying(36) NOT NULL,
+                    -- These are found in "properties" of an alert Feature:
+                    alert_type text,
+                    area_alert_ha double precision,
```
rudokemper, line 413:
```suggestion
                    area_alert_ha double precision,  -- only present for polygon
```

rudokemper (approved):

f/frizzle/alerts/alerts_gcs.py
IamJeffG: Will do it in a separate PR

IamJeffG: https://github.com/ConservationMetrics/gc-scripts-hub/pull/38

PR #37, opened 2024-12-12
Remove SQL injection vectors

IamJeffG:
## Goal

The code had SQL injection vulnerabilities from interpolating {table_name} (and in some cases, column names) directly into the query string.

## What I changed

To address this we use the `psycopg2.sql` module, which provides safe ways to compose SQL queries dynamically

I also replaced the entire `_table_exists()` method with `CREATE TABLE IF NOT EXISTS`, which reduced the number of places I had to deal with constructing elaborate SQL queries.

### DIFF
--- a/f/frizzle/comapeo/comapeo_observations.py
+++ b/f/frizzle/comapeo/comapeo_observations.py
@@ -2,6 +2,7 @@
 # psycopg2-binary
 # requests~=2.32
 
+import json
 import logging
 import mimetypes
 import os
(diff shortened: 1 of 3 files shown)


f/frizzle/comapeo/comapeo_observations.py
```diff
@@ -402,19 +419,16 @@ def handle_output(self, outputs):
...
                 try:
-                    # Wrap column names in single quotes
-                    columns = ", ".join(f'"{name}"' for name in row.keys())
-
-                    # Prepare the values list, converting None to 'NULL'
-                    values = [
-                        f"{value}" if value is not None else None
-                        for value in row.values()
-                    ]
+                    cols, vals = zip(*row.items())
```
IamJeffG, line 422: From this line to the rest of the file could use extra attention by reviewer.  In particular the handling of non-primitive (list and dict) values.

rudokemper: Looks fine to me. And I tested the code with real CoMapeo data to be sure.

rudokemper (approved):
Thanks!

PR #38, opened 2024-12-13
Add "confidence" to alerts table

IamJeffG:
## What I changed

1. Add confidence to "alerts" SQL table.
    - `smallint` as the only values I've seen in the wild are {0, 1}
    - @rudokemper do you know what this column means?  Is there reason to believe it could contain values that aren't smallint?

2. Change datatype for "alerts__metadata.confidence" column.
    - `smallint` as the only values I've seen in the wild are {0, 1}.
        - Therefore changed the typecast from `float` to `int`.
    - @rudokemper do you know what this column means?  Is there reason to believe it could contain values that aren't smallint?

## What I'm not doing here

Tests for different values of these columns.

### DIFF
--- a/f/frizzle/alerts/alerts_gcs.py
+++ b/f/frizzle/alerts/alerts_gcs.py
@@ -377,7 +377,7 @@ class AlertsDBWriter:
                     year text NOT NULL,
                     total_alerts text NOT NULL,
                     description_alerts text,
-                    confidence smallint,
+                    confidence real,
                     metadata_uuid text,
                     source text,
                     alert_source text
@@ -396,7 +396,7 @@ class AlertsDBWriter:
                 alert_type text,
                 area_alert_ha double precision,  -- only present for polygon
                 basin_id bigint,
-                confidence smallint,
+                confidence real,
                 count bigint,
                 date_end_t0 text,
                 date_end_t1 text,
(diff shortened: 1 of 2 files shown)


rudokemper (approved):
I confirmed with the alerts provider that the values for `confidence` will only ever be 0 or 1.

rudokemper:
...Actually, the alerts provider is hedging; they are not sure if they will someday have a `0.5` value. So it is better off as a float or decimal.

IamJeffG:
Bigger question: What does this value _mean_?  I assume it's not as simple as "it's a somewhat opaque value emitted by an ML model" or else it wouldn't be only 0, 0.5, or 1...

Here's what I'm getting at:
If we don't (or are not supposed to ) know the semantics of this field, why should we bother exposing it in the warehouse?

Are we just adding noise to the database tables? Data that is not actionable?

IamJeffG:
> but for computed values like these that were only used for processing, I think it's a good idea.

Sorry, please clarify: What action do you think is a good idea for computed values used only for processing?

And thanks for these answers and thoughts!  Ok, sounds like `confidence` is meaningful. I will keep it in the DB and also make it (in both tables) a float.

rudokemper:
> What action do you think is a good idea for computed values used only for processing?

Omit them (`grid`, `label`, and `count`) from being written to the db.

Issue #39, opened 2024-12-13
[frizzle] Discussion: consider renaming "frizzle"


rudokemper:
Until now, we have used the name "frizzle" (referring to [the character of Ms. Frizzle from The Magic School Bus TV Series](https://en.wikipedia.org/wiki/The_Magic_School_Bus_(TV_series))) to refer to our data pipeline tooling. I believe this follows a long-standing pattern at CMI to use fun names for internal tooling, but I might be wrong.

"Frizzle" was first used to describe a unique implementation of the asynchronous actor model concurrency library [thespian](https://thespianpy.com/) together with Flask, and a few actor components for continuous polling of data from various data sources (Kobo, CoMapeo, Alerts). Next, it was used to describe our specific deployment of [dagster](https://dagster.io/) together with an admin UI built in Nuxt.js, and the same data ETL operations but rewritten as Dagster op graphs. 

Today, "frizzle" is referring exclusively to a set of scripts for continuous data ETL on Windmill, i.e. what lives in `/f/frizzle`.

In a way, it makes sense to persist the name. These same scripts were the ones being operationalized in `frizzle-thespian` and `frizzle-dagster` (and the very _raison d'etre_ for these tools), and more than anything, the scripts represent the core product of what CMI is building for data integration.

However, I think our use of Windmill as a general scripting library where "frizzle" is exposed to the user and exists alongside other categories of scripts, complicates this notion. 

Previously, the name "frizzle" was entirely hidden from the end user - for example, in the now-scrapped Nuxt.js `admin_ui/` work in `frizzle-dagster` I was opting for a name "GuardianConnector Data Connections" to present something more meaningful to the user than "frizzle". Now, in Windmill, for the first time our users can encounter the name "frizzle" - it is exposed in the Windmill UI as the name of the script folder in many places.

![image](https://github.com/user-attachments/assets/a8eed69b-1f4b-4c12-aaab-df8a1642a345)

Personally, I have never watched the TV show from which the name came, and I have needed to Google it a few times to explain what it refers to to explain what it is of our collaborators and users. I reckon that the name "frizzle" will not ring a bell for most GuardianConnector users. I would even be surprised if _any_ of them know the name.

With the imminent merge of a PR to add a `f/export` scripts directory, I wonder if we're better off calling a spade a spade, and just rename "frizzle" to an explicit name that makes sense to our users and differentiates it in a logical way from the other scripts directories. 

What comes to mind:

[...]

PR #40, opened 2024-12-16
Rename frizzle to connectors

rudokemper:
## Goal

To opt for a semantically meaningful name for our data connectors instead of "frizzle", per https://github.com/ConservationMetrics/gc-scripts-hub/issues/39.

After merging this, I will need to update schedules on all of our production deployments.

## What I changed

Renamed `frizzle` to `connectors`

## What I'm not doing here

Worry about the use of "frizzle" in our path naming conventions. That is a broader issue across the stack and mostly hidden from the (imagined future GC admin) user. The exception being that they will see it as a default value for `attachment_root` or `destination_path` in some of the scripts, but that's not a big issue.


IamJeffG (approved):
Heads up that after merging & syncing these changes to a windmill deployment, scheduled runs will need to be deleted & re-created (script name changed)

PR #41, opened 2024-12-16
Comapeo cleanup

### LINKED ISSUE #26: [frizzle/comapeo] CoMapeo script improvements and cleanup
## Feature Request

A few improvements needed:

### Renamed keys should be unique and <= 63 chars in length

Scanning the [comapeo-schema](https://github.com/digidem/comapeo-schema), it does not appear that any key formatting or char limits are enforced.

Therefore, when [renaming and snake-casing keys for db column writing](https://github.com/ConservationMetrics/gc-scripts-hub/blob/55c24ca821840ad5f71f68df420feb52e4feb243/f/frizzle/comapeo/comapeo_observations.py#L388-L392) we should guarantee that keys are always unique and <= 63 chars in length (to abide by Postgres column limits).

The work done in the [kobotoolbox_responses script](https://github.com/ConservationMetrics/gc-scripts-hub/blob/55c24ca821840ad5f71f68df420feb52e4feb243/f/frizzle/kobotoolbox/kobotoolbox_responses.py#L124-L161) will surely be helpful here.

### `g__type` column

Let's create a `g__type` column and detect geometry based on `g__coordinates`, since CoMapeo doesn't currently provide that.

### Reassign new `observations` dictionary back to the list

`observation` rewrites from line 211 onward are not getting reassigned back to the list.

rudokemper:
## Goal

Closes #26. Handles a few outstanding issues in the CoMapeo script.

## What I changed

1. Bugfix: reassigned transformed observations back to the list in `download_and_transform_comapeo_data`; I think I missed this in rewriting the script from `frizzle-dagster`.
2. Added a missing `g__type` column (with a comment that we may need to account for future geometries from the CoMapeo API, the schema for which is unknown)
3. Instead of only snake casing keys, added a `normalize_and_snakecase_keys` function to normalize keys in a more robust way with potential key collision and character limit handling (for Postgres column limits).
4. Expanded tests: unit test for `normalize_and_snakecase_keys`, and in the e2e tests added a few assertions to actually ensure that rows are being written in the expected way. (To wit, tests were missing the bug in (1))

### DIFF
--- a/f/frizzle/comapeo/comapeo_observations.py
+++ b/f/frizzle/comapeo/comapeo_observations.py
@@ -232,7 +232,7 @@ def download_and_transform_comapeo_data(
             logger.info("Response received: ", response.text)
             raise ValueError("Invalid JSON response received from server.")
 
-        for observation in current_project_data:
+        for i, observation in enumerate(current_project_data):
             observation["project_name"] = project_name
             observation["project_id"] = project_id
 
@@ -284,6 +284,8 @@ def download_and_transform_comapeo_data(
                             attachment_failed = True
 
                 observation["attachments"] = ", ".join(filenames)
+
+            current_project_data[i] = observation
 
         # Store observations in a dictionary with project_id as key
         comapeo_data[final_project_name] = current_project_data
--- a/f/frizzle/comapeo/comapeo_observations.py
+++ b/f/frizzle/comapeo/comapeo_observations.py
@@ -172,35 +172,44 @@ def download_attachment(url, headers, save_path):
         return None
 
 
-def normalize_and_snakecase_keys(dictionary):
+def normalize_and_snakecase_keys(dictionary, special_case_keys=None):
     """
     Converts the keys of a dictionary from camelCase to snake_case, handling key collisions and truncating long keys.
+    Optionally leaves specified keys unchanged.
 
     Parameters
     ----------
     dictionary : dict
         The dictionary whose keys are to be converted.
+    special_case_keys : set, optional
+        A set of keys that should not be converted.
 
     Returns
     -------
     dict
         A new dictionary with the keys converted to snake_case and truncated if necessary.
     """
+    if special_case_keys is None:
+        special_case_keys = set()
+
     new_dict = {}
     items = list(dictionary.items())
     for key, value in items:
-        new_key = re.sub("([a-z0-9])([A-Z])", r"\1_\2", key).replace("-", "_").lower()
+        if key in special_case_keys:
+            final_key = key
+        else:
+            new_key = (
+                re.sub("([a-z0-9])([A-Z])", r"\1_\2", key).replace("-", "_").lower()
+            )
+            base_key = new_key[:61] if len(new_key) > 63 else new_key
+            final_key = base_key
+            if len(new_key) > 63:
+                final_key = f"{base_key}_1"
 
-        base_key = new_key[:61] if len(new_key) > 63 else new_key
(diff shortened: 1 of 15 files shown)


f/frizzle/comapeo/tests/comapeo_observations_test.py
```diff
@@ -25,11 +51,23 @@ def test_script_e2e(comapeoserver, pg_database, tmp_path):
...
-        # Survey responses from forest_expedition are written to a SQL Table
+        # Survey responses from forest_expedition are written to a SQL Table in expected format
         with conn.cursor() as cursor:
             cursor.execute("SELECT COUNT(*) FROM comapeo_forest_expedition")
             assert cursor.fetchone()[0] == 3
 
+            cursor.execute(
+                "SELECT column_name FROM information_schema.columns WHERE table_name = 'comapeo_forest_expedition'"
+            )
+            columns = [row[0] for row in cursor.fetchall()]
```
IamJeffG, line 62:
```suggestion
            cursor.execute("SELECT * FROM comapeo_forest_expedition LIMIT 0")
            columns = [desc[0] for desc in cursor.description]
```

I want to get out of the habit of querying `information_schema`: it's a leaky abstraction.

f/frizzle/comapeo/comapeo_observations.py
```diff
@@ -172,8 +172,38 @@ def download_attachment(url, headers, save_path):
         return None
 
 
-def camel_to_snake(name):
-    return re.sub("([a-z0-9])([A-Z])", r"\1_\2", name).lower()
+def normalize_and_snakecase_keys(dictionary):
+    """
+    Converts the keys of a dictionary from camelCase to snake_case, handling key collisions and truncating long keys.
```
IamJeffG, line 177: Some more test cases for you to add to the tests. I was pleasantly surprised to see these working!

```py
collisions = {
    "abcDef": 1,
    "abc-def": 2,
    "AbcDef": 3
}

newdict = normalize_and_snakecase_keys(collisions)
assert newdict == {'abc_def': 1, 'abc_def_2': 2, 'abc_def_3': 3}
```

f/frizzle/comapeo/comapeo_observations.py
```diff
@@ -232,28 +262,27 @@ def download_and_transform_comapeo_data(
...
-                for k, v in observation.items()
-            }
+            # Create k/v pairs for each tag
+            if "tags" in observation:
+                for key, value in observation["tags"].items():
+                    observation[key] = value
+                del observation["tags"]
+
+            # Convert keys from camelCase to snake_case, handling key collisions and char limits
+            observation = normalize_and_snakecase_keys(observation)
```
IamJeffG, line 276: Previously we special-cased `docId` right here, never transforming it. That way the key was sure to be as-expected when we check it in line 407 (of the old code).

By making this change, we risk `docId` being a key that gets modified by `normalize_and_snakecase_keys`:

```py

observation = {"DOC-id": "phony", "docId": "the-real-one"}
normalize_and_snakecase_keys(observation)
>>> {'doc_id': 'phony', 'doc_id_2': 'the-real-one'}
```

IamJeffG (approved):

f/frizzle/comapeo/comapeo_observations.py
rudokemper: Good catch, thanks. Opted for adding a special-case param inside `normalize_and_snakecase_keys`.

PR #43, opened 2024-12-18
Refactor `alerts_gcs` to resolve a number of bugs and improve readability

### LINKED ISSUE #21: [connectors/alerts] Improve alerts metadata handling
3 improvements are needed for processing alerts metadata in the `alerts_gcs` script:

- [x] Currently, the script is saving the entire `alerts_metadata.csv` file from GCP to `destination_path` along with the TIF and GeoJSON files, which is not desired behavior.
- [x] Additionally, the script seems to be duplicating rows from `alerts_metadata.csv` instead of updating existing rows. 
- [x] Lastly, a small fix: for the `confidence` column, there are NaN values for rows where no value is found in the CSV. If the alerts provider does not provide a value for a given month, the value should stay null.

Once done and merged, let's truncate all of the `alerts__metadata` tables for our prod databases, and rerun alerts processing.

### LINKED ISSUE #27: [connectors/alerts] store vector files in the same dir as raster
## Feature Request

The `alerts_gcs` script downloads both vector and raster data, but stores them in different directories:

**Vector:**  `/datalake/change_detection/alerts/{territory_id}/vector/{year}/{month}/{alert_id}.geojson`

**Raster:**  
`/datalake/change_detection/alerts/{territory_id}/{year}/{month}/{alert_id}/images/...`

Let's store the vector GeoJSON in the same path as raster, namely:
`/datalake/change_detection/alerts/{territory_id}/{year}/{month}/{alert_id}/{alert_id}.geojson`

In doing so, let's make sure the script README is accurate.

rudokemper:
## Goal

Closes #21. Closes #27.

For this PR, I had the intention to stay scoped on the work described in these two issues. However, in working on the tasks in #21 in particular, I found myself wanting to improve substantial aspects of the script including overall readability and separation of concerns. 

So, this ended up being a much larger refactor of the script. If the reader wants me to break this up into smaller PRs, I can do so, but I hope my notes in the following section suffice for reviewing this.

## What I changed

**`sync_gcs_to_local`**

* Don't add `alerts_metadata` to the list of files to download to storage. Instead, retrieve the blob, but store the contents in memory for later processing.
* Use the same `rel_filepath` for GeoJSON and TIFFs per the `_get_rel_filepath` helper function, which appends `images/` only if it's an image, thus solving #27.
* Track what files are actually downloaded using  `downloaded_files` (Set) instead of just returning `files_to_download`.
* After downloading, create separate lists of GeoJSON and TIFF files from `downloaded_files`, and return these alongside `alerts_metadata`

**`process_files`**

I broke up `process_files` into smaller functions: `convert_tiffs_to_jpg`, and `prepare_alerts_metadata`:

* `convert_tiffs_to_jpg`: This function receives the list of TIFF files as an argument and does the conversion to JPG work. 
* `prepare_alerts_metadata`: This function reads the `alerts_metadata` CSV and uses the same Pandas handling to prepare dictionary as before. A few meaningful additional changes:
  * Before, if a value was missing in the CSV, the entry would be set as NaN, which made its way all the way to the database record, as described in #21. Now, I am using `numpy.nan` to convert those values to None. (I couldn't find a way to do this without using `numpy`, suggestions welcome if a more parsimonious method is known.)
  * We were generating a UUID but solely based on the alerts_metadata filepath, hence the value was the same for each record and therefore pretty meaningless. Instead, I added a new helper function `_generate_uuid` which uses  `sha256sum` to generate a unique UUID based on row content. This will help us later when writing to the database and avoiding duplications...

**`prepare_alerts_data`** 

Does some of the same work processing the GeoJSON files and appending a few fields as was found in `load_alerts_to_postgis`.

**`AlertsDBWriter`**

* Pass cursor to the methods to create tables, instead of creating a new one (as noted in a FIXME comment)
* Set primary keys for each `alerts` and `alerts__metadata` table. The absence of these led to duplicate writes as described in #21.
* For `alerts__metadata`, set more accurate column types. And remember that unique UUID that we are now generating? This is now being set as the `_id` (PRIMARY KEY) column, so that we can check for uniqueness when writing.
* Introduce the same `_safe_insert` method to avoid SQL injection that was introduced in the other connector scripts.

> [!IMPORTANT]
> The work here introduces breaking changes. We will have to drop and recreate tables on all of our databases. But this change is needed to avoid mass duplication of alerts, which is the status quo!

**Tests**

* Adapted paths and assertions in accordance to new expected paths.
* In `alerts_history.csv`,
  * did not set `confidence` for one row, so as to assert that it will be None (and not "NaN" as before) when written to the database
  * added one additional row which is a duplicate of a previous row, so as to assert that it was not written to the database due to `UniqueViolation` error.
* Test to make sure `alerts_history.csv` is not saved to disk (possibly not needed...)

**Other**

* Substituted `pathlib.Path` in place of `os.path`.
* Across the board, I improved our variable naming conventions, removed unused arguments, added docstrings, spell-checked and removed redundant comments.


f/connectors/alerts/tests/alerts_gcs_test.py
```diff
@@ -85,3 +95,6 @@ def test_script_e2e(pg_database, mock_alerts_storage_client, tmp_path):
         assert (
             asset_storage / "100/2023/09/202309900112345671/images" / f"{basename}.jpg"
         ).exists()
+
+    # Alerts metadata is not saved to disk
+    assert not (asset_storage / "alerts_history.csv").exists()
```
rudokemper, line 100: Arguably, this assertion can be omitted, since there is no reason to expect this moving forward. But I thought to keep it to ensure that we don't accidentally save this CSV in the future, since it has already happened.

f/connectors/alerts/alerts_gcs.py
```diff
@@ -124,221 +131,251 @@ def _get_tif_rel_filepath(blob_name, local_file_path, territory_id):
...
+    filtered_df = df[df["territory_id"] == territory_id].copy()
+
+    # Generate a unique UUID for each row
+    filtered_df["metadata_uuid"] = filtered_df.apply(_generate_uuid, axis=1)
+
+    # TODO: Currently, this script is only used for Terras alerts. Let's discuss a more sustainable approach with the alerts provider(s).
+    filtered_df.loc[:, "alert_source"] = "terras"
+
+    # Replace all NaN values with None
+    filtered_df = filtered_df.replace({nan: None})
```
rudokemper, line 333: Is there a way to do this without `numpy.nan`?

f/connectors/alerts/alerts_gcs.py
```diff
@@ -480,116 +546,137 @@ def handle_output(self, geojsons, metadatas):
...
+                    _id = metadata.get("metadata_uuid")
                     territory_id = metadata.get("territory_id")
                     type_alert = metadata.get("type_alert")
                     month = metadata.get("month")
                     year = metadata.get("year")
                     total_alerts = metadata.get("total_alerts")
                     description_alerts = metadata.get("description_alerts")
                     confidence = metadata.get("confidence")
-                    metadata_uuid = metadata.get("metadata_uuid")
-                    source = metadata.get("source")
```
rudokemper, line 552: FYI: this column was storing the name of the CSV e.g. `alerts_history.csv`. I don't think it's necessary to include this.

f/connectors/alerts/alerts_gcs.py
```diff
@@ -124,221 +131,251 @@ def _get_tif_rel_filepath(blob_name, local_file_path, territory_id):
...
-        An additional file to download: mean tto be the filename containing alerts metadata.
+        An additional file to sync from the GCS bucket.
 
     Returns
     -------
-    set of str
-        the (local) file names that were written.
+    tuple
+        A tuple containing a set of str with the local file names that were written
+        and the alerts metadata content.
```
IamJeffG, line 172: While it's correct that it's a tuple, numpy doc convention would split out the multiple return elements and document each:

```suggestion
    geojson_files : set of str
        local file names of the geojson files that were written
    tiff_files : set of str
        local file names of the TIFF files that were written
    alerts_metadata : str
        alerts metadata content.
```

f/connectors/alerts/alerts_gcs.py
```diff
@@ -124,221 +131,251 @@ def _get_tif_rel_filepath(blob_name, local_file_path, territory_id):
...
+    bucket_name,
+    territory_id,
+    alerts_metadata_filename,
 ):
     """Download files from a GCS bucket to a local directory.
 
     Parameters
     ----------
-    dst_path : str
+    destination_path : str
```
IamJeffG, line 158: I see you coercing `destination_path` (str) to a Path many times in this function.  You could do this once at the very start.


You could even make the function a tad more flexible to accept either str or Path; I see this done in many other libraries.  Example:

```py
def sync_gcs_to_local(destination_path):
     """
    Parameters
    ------------
    destination_path : str or pathlib.Path
        The path to the destination. Can be a string or a pathlib.Path object.
    """
    destination_path = Path(destination_path)
    ...
```

This works because the Path constructor can take either a string or another Path object as an argument. If you pass another Path object, it will simply return that object unchanged.

f/connectors/alerts/alerts_gcs.py
```diff
@@ -124,221 +131,251 @@ def _get_tif_rel_filepath(blob_name, local_file_path, territory_id):
...
+        blob.download_to_filename(rel_filepath / filename)
 
-        if not os.path.exists(rel_filepath):
-            os.makedirs(rel_filepath)
+        downloaded_files.add(str(rel_filepath / filename))
 
-        blob.download_to_filename(os.path.join(rel_filepath, filename))
+    # Create lists of GeoJSON and TIFF files
+    geojson_files = [f for f in downloaded_files if f.endswith(".geojson")]
+    tiff_files = [f for f in downloaded_files if f.endswith((".tif", ".tiff"))]
```
IamJeffG, line 217: Do you find it strange that this function might download a file and then not return its name?  (This would happen if GCS contains a file that's not geojson or tiff).

Tempted to suggest that this check go BEFORE you download, and not download files that don't match?

IamJeffG: Also the old code's extension comparison was case-insensitive, but your new code's is case-sensitive.

f/connectors/alerts/alerts_gcs.py
```diff
@@ -124,221 +131,251 @@ def _get_tif_rel_filepath(blob_name, local_file_path, territory_id):
...
+    df = pd.read_csv(StringIO(alerts_metadata))
+
+    # Filter DataFrame based on territory_id
+    filtered_df = df[df["territory_id"] == territory_id].copy()
+
+    # Generate a unique UUID for each row
+    filtered_df["metadata_uuid"] = filtered_df.apply(_generate_uuid, axis=1)
+
+    # TODO: Currently, this script is only used for Terras alerts. Let's discuss a more sustainable approach with the alerts provider(s).
+    filtered_df.loc[:, "alert_source"] = "terras"
```
IamJeffG, line 330: Any reason you changed this from...?

    filtered_df["alert_source"] = "terras"


I could imagine you had a reason like perhaps Pandas was giving a warning the old way.  But I doubt it. The old way is what I see all the time (and use myself), and while I believe the new way is correct, it feels ... opaque.

f/connectors/alerts/alerts_gcs.py
```diff
@@ -124,221 +131,251 @@ def _get_tif_rel_filepath(blob_name, local_file_path, territory_id):
...
-    outputs["alerts_metadata"] = processed["alerts_metadata"]
-    outputs["geojsons"] = []
-
-    for file_name in processed["geojsons"]:
-        if os.path.exists(os.path.join(local_directory, file_name)):
+    # Convert CSV bytes to DataFrame
+    df = pd.read_csv(StringIO(alerts_metadata))
+
+    # Filter DataFrame based on territory_id
+    filtered_df = df[df["territory_id"] == territory_id].copy()
```
IamJeffG, line 324: nit: You should not need a copy.  Does the following work without warning?

```suggestion
    filtered_df = df.loc[df["territory_id"] == territory_id]
```

I guess maybe you are worried about accidentally mutating the original `df` later in line 330? But nobody outside this function has a reference to it.

f/connectors/alerts/alerts_gcs.py
IamJeffG:
```suggestion
    filtered_df = filtered_df.fillna(None)
```

f/connectors/alerts/alerts_gcs.py
```diff
@@ -124,221 +131,251 @@ def _get_tif_rel_filepath(blob_name, local_file_path, territory_id):
...
+
+    # Generate a unique UUID for each row
+    filtered_df["metadata_uuid"] = filtered_df.apply(_generate_uuid, axis=1)
+
```
IamJeffG, line 328: the old code also added "_geolocation"  and "source" columns.  Do we need that? 


(I am not sure what that was for)

f/connectors/alerts/alerts_gcs.py
```diff
@@ -124,221 +131,251 @@ def _get_tif_rel_filepath(blob_name, local_file_path, territory_id):
...
+    # Replace all NaN values with None
+    filtered_df = filtered_df.replace({nan: None})
+
+    # Convert DataFrame to list of dictionaries
+    prepared_alerts_metadata = filtered_df.to_dict("records")
+
+    logger.info("Successfully prepared alerts metadata.")
+
+    return prepared_alerts_metadata
```
IamJeffG, line 340: the old code also added "_geolocation"  and "source" columns.  Do we need that? 


(I am not sure what that was for)

f/connectors/alerts/alerts_gcs.py
```diff
@@ -124,221 +131,251 @@ def _get_tif_rel_filepath(blob_name, local_file_path, territory_id):
...
-      and generates a unique metadata UUID for the filtered CSV records, returning them.
+
+    def sha256sum(data):
+        h = hashlib.sha256()
+        h.update(data.encode("utf-8"))
+        return h.hexdigest()
+
+    row_data = json.dumps(row.to_dict(), sort_keys=True)
+    row_hash = sha256sum(row_data)
+    return str(uuid.uuid5(uuid.NAMESPACE_DNS, row_hash))
```
IamJeffG, line 294: I have some doubts about this new `_generate_uuid` function.

However before I get into them, can you explain (ideally in some code comments and point me to them) what the point of generating a UUID is for?  I see the PR description refer to avoiding duplicates but help me understand the bigger picture please.

That way I can be sure to give feedback that's relevant and not misguided.

IamJeffG (changes requested):
I stopped reviewing at the end of `prepare_alerts_metadata`.

Honestly this PR is doing too much.  If it's easy, do you think we could split the changes to `AlertsDBWriter` to a separate PR? 

(I have not looked at your changes to `AlertsDBWriter` yet, but already I feel some hesitation: There are a number of psycopg best practices that we fail to meet throughout this repo, and on top of that it will be best to move to psycopg3.  I really hesitate to change much in AlertsDBWriter without making the same changes & upgrades to all our scripts.)

f/connectors/alerts/alerts_gcs.py
rudokemper: No. `_geolocation` was a superfluous (I assume accidental) addition. I thought I had clarified somewhere that `source` stored the name of the filename e.g. "alerts_history.csv" which I do not think is necessary to store.

f/connectors/alerts/alerts_gcs.py
rudokemper: See other comment.

f/connectors/alerts/alerts_gcs.py
rudokemper: https://github.com/ConservationMetrics/gc-scripts-hub/blob/822d2ea52ef11baaad584484b3b0ec181d2522d0/f/connectors/alerts/alerts_gcs.py#L410 creates a PRIMARY_ID column `_id`. Previously, there were no PRIMARY KEY columns and hence alerts__metadata rows were being duplicated every time this script is run.

https://github.com/ConservationMetrics/gc-scripts-hub/blob/822d2ea52ef11baaad584484b3b0ec181d2522d0/f/connectors/alerts/alerts_gcs.py#L632 sets the value of `_id` as the row's UUID. And the reason we generate a UUID to begin with is because we don't get any unique identifier from the `alerts_history.csv` file otherwise.

f/connectors/alerts/alerts_gcs.py
rudokemper: This does not seem to work: "ValueError: Must specify a fill 'value' or 'method'."

For posterity, I also tried `filtered_df = filtered_df.where(pd.notna(filtered_df), None)` which did not seem to work.

rudokemper:
> Honestly this PR is doing too much.

Yeah, that's fair. As you can tell I was conscious of that fact in my PR description. I'll do better the next time; I didn't anticipate that I would change so much at the outset.

> If it's easy, do you think we could split the changes to AlertsDBWriter to a separate PR?
>
> (I have not looked at your changes to AlertsDBWriter yet, but already I feel some hesitation: There are a number of psycopg best practices that we fail to meet throughout this repo, and on top of that it will be best to move to psycopg3. I really hesitate to change much in AlertsDBWriter without making the same changes & upgrades to all our scripts.)****

Yeah, I know you have expressed this desire to move to psycopg3 a number of times before.

The main reason the work here looks like a lot is because (1) I implemented the `_safe_insert` method introduced in a different script, and (2) am passing a cursor to class methods instead of instantiating a new one each time, as was suggested in a `FIXME` comment. Otherwise, I'm mainly just renaming variables and adapting the table schemas (which is required for this PR to work to meet the aims of #21). If you'd like, I can revert all changes with the exception of the necessary ones.

f/connectors/alerts/alerts_gcs.py
rudokemper: Honestly, I was getting an error and remember needing to change this, but that is no longer the case, and I couldn't tell you why. So I reverted to the original.

rudokemper:
I have gone ahead and reverted my changes to `AlertsDBWriter` so as to leave it untouched apart from variable + function name changes, db table schema changes, and minor cleanup.

f/connectors/alerts/alerts_gcs.py
IamJeffG: you are right.  If your goal is to not `import numpy`, you can do this:

`filtered_df.replace({float("nan"): None})`


`nan` is how pandas represents missing data, and it's efficient in a column whose datatype is float.  Replacing  `nan` to `None` will upgrade that column's dtype to `object`.  In this case it's not a problem, but won't always be what you want.

f/connectors/alerts/alerts_gcs.py
```diff
@@ -124,221 +131,264 @@ def _get_tif_rel_filepath(blob_name, local_file_path, territory_id):
...
-
-    for file_name in processed["geojsons"]:
-        if os.path.exists(os.path.join(local_directory, file_name)):
+    # Convert CSV bytes to DataFrame
+    df = pd.read_csv(StringIO(alerts_metadata))
+
+    # Filter DataFrame based on territory_id
+    filtered_df = df.loc[df["territory_id"] == territory_id]
+
+    # Generate a unique UUID for each row
```
IamJeffG, line 339:
```suggestion
    # Hash each row into a unique UUID.
    # We don't get any unique identifier from the alerts_history.csv file otherwise.
```

from your response above.

f/connectors/alerts/alerts_gcs.py
IamJeffG: Thanks for explaining.

There is no _great_ solution here, every solution I think of is prone to edge cases, such as what datatype pandas uses to represent a column, or the order of the columns, etc.

My first suggestion is to consider whether you really want to consider all columns of the row, or only one or two "index columns." I'd think that [territory_id, month,  year] might be enough?  Or [territory_id, month,  year, alert_source, description_alerts].  That way if for whatever reason there was a change in some other field like "confidence" (only), we don't insert another row.  I'd consider that a duplicate.

Second, JSON-ifying is not great, but OTOH nothing is great.  Problems with hashing JSON is that the hash depends on non-information-carrying elements of the serialization, for example whitespace and key ordering. That said, I think your implementation should not change the hash as long as nothing else changes inside the `json` lib.  I considered (and I guess slightly prefer) just using Panda's own [`pandas.util.hash_pandas_object`] (https://pandas.pydata.org/docs/reference/api/pandas.util.hash_pandas_object.html) to return one hash value for reach row of the dataframe.  Now, it has similar problems such as depending on column order and on index value.  You can somewhat achieve a similar result, in much less code, using:

```py
    filtered_df["metadata_uuid"] = pd.util.hash_pandas_object(filtered_df.sort_index(axis=1), index=False)
    #optionally wrap this in uuid.uuid5(...)
```

Finally, consider choosing the most appropriate postgresql datatype for `_id`: If it's really a uuid then use `uuid`, if it's an integer (as with hash_pandas_object) then choose `bigint`.

All this is food for thought.  As I said, I am not in love with any solution, so it's also fine if you do not touch this.

f/connectors/alerts/alerts_gcs.py
```diff
@@ -516,69 +567,57 @@ def handle_output(self, geojsons, metadatas):
 
                         except errors.UniqueViolation:
                             logger.info(
-                                f"Skipping insert due to UniqueViolation, this alert has been processed already in the past: {source}"
+                                f"Skipping insert due to UniqueViolation, this entry has been processed already in the past: {_id}"
```
IamJeffG, line 570: If I follow, this never would have triggered before your change. There was not a unique constraint before this.

f/connectors/alerts/alerts_gcs.py
```diff
@@ -423,37 +472,41 @@ def _create_alerts_table(self, table_name):
...
+        alerts_metadata (list): A list of metadata objects associated with the alerts, containing additional information to be stored in a separate metadata table.
         """
         table_name = self.table_name
         conn = self._get_conn()
         cursor = conn.cursor()
 
         try:
-            if geojsons:
-                # check to see if we need to create the table for the first time
-                # FIXME: this func creates its own cursor but we already have one
```
IamJeffG, line 437: yeah, cursors are extremely lightweight and it's common to create new ones all the time.  I see no problem with various functions creating (totally managing) their own.

IamJeffG (approved):
thanks for sticking with me on this!

f/connectors/alerts/alerts_gcs.py
rudokemper: Yes, that is correct! This was an oversight already in the frizzle-dagster implementation, resulting in https://github.com/ConservationMetrics/frizzle-dagster/issues/139

IamJeffG:
> I have gone ahead and reverted my changes to AlertsDBWriter so as to leave it untouched apart from variable + function name changes, db table schema changes, and minor cleanup.

Thanks for doing this! Helped a lot!

f/connectors/alerts/alerts_gcs.py
rudokemper: Thanks for this! I didn't know about `hash_pandas_object`. Despite its own shortcomings, I find that using it nicely streamlines this function where we can now accomplish everything we want with Pandas. 

I'm about to push a commit that implements this change; I also worked in your suggestion to only hash based on the most important columns, and a comment in `alerts_gcs_test` to show how we are testing all of this.

Issue #20, opened 2024-12-09
[connectors/kobotoolbox] Clean up and improve `kobo_responses.py` script


rudokemper:
The `kobo_responses.py` script could use some refactoring to be cleaned up and streamlined post-Dagster port, similar to what's been done for CoMapeo. In so doing, let's:

- [x] add docstrings to document functions
- [x] remove unused vars
- [x] use `_geolocation` fields to construct `g__type` and `g__coordinates`
- [x] better sanitizing so that letters with diacritics are not erased but rewritten as the base letter without marking

PR #45, opened 2024-12-20
For KoboToolbox, construct `g__` fields from `_geolocation` field

rudokemper:
## Goal

Toward https://github.com/ConservationMetrics/gc-scripts-hub/issues/20. This is stacked on PR https://github.com/ConservationMetrics/gc-scripts-hub/pull/44.

## What I changed

* Added a `transform_form_data` function where we can make transformations to the form data as needed.
* Currently, I am only using this to do one thing: take `_geolocation` to construct `g__type` and `g__coordinates` fields (this seemingly was forgotten in the transition from frizzle-thespian to frizzle-dagster).
* In the future we can also use this function to remove any unwanted fields e.g. meta content.

## What I'm not doing here

I went down a rabbit hole of investigating whether `_geolocation` is ever anything other than a Point. There are polygon and line field options in KoboToolbox but these are never set as `_geolocation`. So I kept things simple and opted not to touch any polygon or line geometry content; it's impossible to know deterministically if these optional fields are ever the intended primary geometry of a given form.

### DIFF
--- a/f/connectors/kobotoolbox/kobotoolbox_responses.py
+++ b/f/connectors/kobotoolbox/kobotoolbox_responses.py
@@ -42,10 +42,10 @@ def main(
         kobo_server_base_url, kobo_api_key, form_id, attachment_root
     )
 
-    prepared_form_data = prepare_form_data(form_data)
+    transformed_form_data = transform_form_data(form_data)
 
     db_writer = KoboDBWriter(conninfo(db), db_table_name)
-    db_writer.handle_output(prepared_form_data)
+    db_writer.handle_output(transformed_form_data)
     logger.info(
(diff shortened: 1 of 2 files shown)


f/connectors/kobotoolbox/kobotoolbox_responses.py
```diff
@@ -145,6 +147,30 @@ def download_form_responses_and_attachments(
     return form_submissions
 
 
+def transform_form_data(form_data):
```
IamJeffG, line 150:
```suggestion
def format_geometry_fields(form_data):
```

or some other name that gives a clue that this is about geometry (as opposed to anything else)

IamJeffG (approved):

PR #47, opened 2024-12-30
Apply `shorten_and_uniqify` to column table

rudokemper:
## Goal

This is stacked on PR https://github.com/ConservationMetrics/gc-scripts-hub/pull/44.

In a recent run for one of our users, I noticed that a very long table name resulted in an error when trying to create the columns mapping table. This is because the `db_table_name` had a length of 58 chars, and currently, we brute append `__columns` to the table name, thus violating Postgres's 63 char limit for table names. 

This PR aims to solve that by applying the `shorten_and_uniqify` function (with a `maxlen=54` argument) to the column table name before appending `__columns`.

## What I changed

* Tweaked the language in `shorten_and_uniqify` to reflect usage for many identifiers, not just keys.
* At the very beginning of the `handle_output` method, create a `columns_table_name` string with the function applied with 54 maxlen and `__column` appended, and use that throughout.
* Added a unit test to test the desired behavior for `__column` rewrites.

## What I'm not doing here

Addressing a case when `db_table_name` itself is > 63 chars! Right now we use this argument as given and don't transform it in any way. But I thought to leave that for a different PR, because the same is true for `alerts` and `comapeo` connectors. Maybe there is a way to set a validation in the `script.yaml` file.

### DIFF
--- a/f/connectors/kobotoolbox/kobotoolbox_responses.py
+++ b/f/connectors/kobotoolbox/kobotoolbox_responses.py
@@ -204,32 +204,32 @@ def _drop_nonsql_chars(s):
     return re.sub(r"[ ./?\[\]\\,<>(){}]", "", s)
 
 
-def _shorten_and_uniqify(key, conflicts, maxlen):
-    """Shorten a key and ensure its uniqueness within a set of conflicts.
+def shorten_and_uniqify(identifier, conflicts, maxlen):
+    """Shorten an identifier and ensure its uniqueness within a set of conflicts.
 
-    This function truncates a key to a specified maximum length and appends a
-    numeric suffix if necessary to ensure uniqueness within a set of conflicting keys.
+    This function truncates an identifier to a specified maximum length and appends a
+    numeric suffix if necessary to ensure uniqueness within a set of conflicting identifiers.
 
     Parameters
     ----------
-    key : str
-        The original key to be shortened and made unique.
+    identifier : str
+        The original identifier to be shortened and made unique.
     conflicts : set
-        A set of keys that the new key must not conflict with.
+        A set of identifiers that the new identifier must not conflict with.
     maxlen : int
-        The maximum allowed length for the key.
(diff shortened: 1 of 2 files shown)


f/connectors/kobotoolbox/kobotoolbox_responses.py
```diff
@@ -469,12 +468,15 @@ def handle_output(self, all_submissions):
         all_submissions : list of dict
         """
         table_name = self.table_name
+        columns_table_name = (
```
IamJeffG, line 471: Add a code comment highlighting that we are NOT using the "uniqify" feature of this function (by passing the empty set of existing names).

Or else don't call that function at all (I'd argue it's a bit misleading to do so) and instead just

```suggestion
        columns_table_name = f"{table_name[:54]}__columns"
```

Note that if you had two very long responses tables with similar prefixes, they might clobber each other's column tables.  It's not the end of the world, as I assume they should nonetheless end up remapped correctly!

IamJeffG (approved):
I wonder if this solution is too clever. Would it be better to add a constraint on the Windmill input that you cannot type names > 54 chars in length? (And add an assertion here to fail if it somehow gets one anyway)

Or maybe do the first anyway, even if you keep these changes.

f/connectors/kobotoolbox/kobotoolbox_responses.py
rudokemper: Ah, much more straightforward, like `slice` in javascript. Great, thanks.

rudokemper:
> Would it be better to add a constraint on the Windmill input that you cannot type names > 54 chars in length?

Fair point! As I mentioned in the PR, we have to add a constraint for table names regardless.

I will still keep the changes so as to only write the `__columns` name once (with your suggestion to slice instead of using the uniqify function), but I'll open a new issue about char limit for table names across the scripts.

PR #52, opened 2024-12-31
Switch alerts `_id` column to UUID type

rudokemper:
## Goal

A real alerts `_id` value for one of our users has 20 characters, which is "out of range" for Postgres BIGINT column (limit: 9223372036854775807, 19 chars). Since this `_id` field is not computationally relevant, it's better off being NUMERIC which has virtually unlimited precision.

### DIFF
--- a/f/connectors/alerts/alerts_gcs.py
+++ b/f/connectors/alerts/alerts_gcs.py
@@ -425,7 +425,7 @@ class AlertsDBWriter:
             query = sql.SQL("""
             CREATE TABLE IF NOT EXISTS {table_name}
             (
-                _id character varying(36) NOT NULL PRIMARY KEY,
+                _id uuid PRIMARY KEY,
                 -- These are found in "properties" of an alert Feature:
                 alert_id text,
                 alert_type text,
--- a/f/connectors/alerts/alerts_gcs.py
+++ b/f/connectors/alerts/alerts_gcs.py
@@ -9,6 +9,7 @@
 
 import json
 import logging
+import uuid
 from io import StringIO
 from pathlib import Path
 
@@ -424,8 +425,9 @@ class AlertsDBWriter:
             query = sql.SQL("""
             CREATE TABLE IF NOT EXISTS {table_name}
             (
-                _id numeric NOT NULL PRIMARY KEY,
+                _id character varying(36) NOT NULL PRIMARY KEY,
                 -- These are found in "properties" of an alert Feature:
+                alert_id text,
                 alert_type text,
                 area_alert_ha double precision,  -- only present for polygon
                 basin_id bigint,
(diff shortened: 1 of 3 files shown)


rudokemper:
Of course, the column could also be set as TEXT, although we have no reason to expect any characters other than numbers.

IamJeffG (commented):
`numeric` is valid but will be slowest and possibly take the most storage of your options.

For simplicity I would go with `text` or `uuid` (the latter is most efficient in time and space, but will require you to do the ol' `uuid5(namespace, the_hash)`  thing)

rudokemper:
@IamJeffG I took your suggestion to implement UUID insofar as it matches what we do for `alerts__metadata`'s primary key.

I wasn't sure what to use for the namespace. I opted for DNS as a standard, to avoid having to generate our own.

f/connectors/alerts/alerts_gcs.py
```diff
@@ -424,8 +425,9 @@ def _create_alerts_table(self, table_name):
             query = sql.SQL("""
             CREATE TABLE IF NOT EXISTS {table_name}
             (
-                _id numeric NOT NULL PRIMARY KEY,
+                _id character varying(36) NOT NULL PRIMARY KEY,
```
IamJeffG, line 428: this would be

`_id uuid PRIMARY KEY`

f/connectors/alerts/tests/alerts_gcs_test.py
```diff
@@ -39,7 +40,7 @@ def _upload_blob(bucket, source_file_name, destination_blob_name):
         "100/raster/2023/09/S2_T1_202309900112345671.tif",
     ]
     for filename in alerts_filenames:
-        source_path = os.path.join(assets_directory, os.path.basename(filename))
+        source_path = Path(assets_directory) / Path(filename).name
```
IamJeffG, line 43: There is (and was before) a super subtle bug here, that would only raise itself if you're running the test suite on a Windows machine.  (At least I think this to be true, and I have seen something like it in the past before)

Basically the `/` operator will, on windows, become a `\`.  That's fine if you're constructing a path for the local filesystem, but this path is not a local file, it's a blob and needs to be `/`

One way (maybe there are others) to do this right would be

```suggestion
        source_path = posixpath.join(assets_directory, Path(filename).name)
```

with an `import posixpath` at top

IamJeffG: Not tested, but you might also have luck with:

```suggestion
        source_path = PosixPath(assets_directory) / PosixPath(filename).name
```

IamJeffG (approved):

f/connectors/alerts/tests/alerts_gcs_test.py
rudokemper: I'll turn this into its own issue since we use this convention in many places in the repo.

IamJeffG: That's fine.  I think most of the places we use this convention, it is correct -- by which I mean we are referring to the local file system, or a network mount that's exposed as a local file system.

This is one of the few (only) cases I'm aware of where we're dealing with blob storage names.

PR #53, opened 2024-12-31
Set default value for `str_replace` in KoboDBWriter class

rudokemper:
## Goal

I noticed in production instances that KoboToolbox questions like `group1/question` are being rewritten as `questiongroup1` columns. The desired format is `question__group1`.

The reason for this is that we are not currently supplying a `str_replace` list of string replacement tuples to `sanitize()`. 

* In frizzle-thespian, we [provided `str_replace` as an optional `FRIZZLE_CONFIG` var](https://github.com/ConservationMetrics/frizzle-thespian/blob/5d9f338be119320c0d7ee97f029a46a9999d9d69/components/warehouse/__init__.py#L129). 
* In frizzle-dagster, we [provided a default value for `str_replace` in the `sanitize()` function itself](https://github.com/ConservationMetrics/frizzle-dagster/blob/cd89c04bf58c819466a074fd8ebb3f13df0bfa73/dagster/core/core/postgres_io.py#L839).

Here, I am opting to align with how we provide a default value for `reverse_properties_separated_by` in the class `__init__` method. This will make it straightforward to pass arguments for these properties in the future if needed.

### DIFF
--- a/f/connectors/kobotoolbox/kobotoolbox_responses.py
+++ b/f/connectors/kobotoolbox/kobotoolbox_responses.py
@@ -309,10 +309,10 @@ class KoboDBWriter:
         automatically handles table width resizing, row-level updates, most data type
         conversions, and sanitizing kobo field names to be SQL-compatible.
 
-        Example of using `reverse_properties_separated_by`:
+        Example of using `reverse_properties_separated_by` and `str_replace`:
         - incoming field => `{"group1/group2/question": "How do you do?"}`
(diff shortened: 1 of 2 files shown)


IamJeffG (approved):
Thanks

I think you can also remove this TODO now:

https://github.com/ConservationMetrics/frizzle-dagster/blob/cd89c04bf58c819466a074fd8ebb3f13df0bfa73/dagster/core/core/postgres_io.py#L838

Issue #48, opened 2024-12-30
[connectors] Set character limit constraint for table names in Windmill


rudokemper:
## Feature Request

Postgres has a 63 character limit for table names, so we should set a constraint for table name arguments across the Windmill scripts. 

For scripts where we add a columns mapping table, we should set a 54 character constraint so that `__columns` can be appended to the table name without violating the limit (see discussion in https://github.com/ConservationMetrics/gc-scripts-hub/pull/47).

And the same for scripts where we add `__metadata` (53 char constraint).

PR #55, opened 2024-12-31
Implement db_table_name character constraints in Windmill and scripts

rudokemper:
## Goal

Closes https://github.com/ConservationMetrics/gc-scripts-hub/issues/48.

## What I changed

* For alerts, add a regex pattern allowing <53 characters (to allow for adding a `__metadata` table), and truncate the metadata table name in the script too, to be safe.
* For kobo, add a regex pattern allowing <54 characters, following up on https://github.com/ConservationMetrics/gc-scripts-hub/pull/47.
* For comapeo, the table writing convention differs since the script can create multiple project tables at once, each prepended by an optional `db_table_prefix`. So for this script, I opted to truncate at 63 chars for each db table to be created.
  * There is a potential edge case here where multiple projects could get truncated to the same name. I didn't want to take the time to address that here, but I left a TODO comment indicating the same.

## Notes

As far as I could tell by looking at other Windmill code and the sparse documentation (https://www.windmill.dev/docs/core_concepts/json_schema_and_parsing#advanced-settings), regex is the only way to do this. There is no `maxLength` or similar setting for String types. The regex validation failure message is a bit ugly on the front end but it works:

image

### DIFF
--- a/f/connectors/alerts/alerts_gcs.script.yaml
+++ b/f/connectors/alerts/alerts_gcs.script.yaml
@@ -29,7 +29,7 @@ schema:
       description: The name of the database table where alerts will be stored.
       default: null
       originalType: string
-      pattern: '^.{0,53}$'
+      pattern: '^.{1,53}$'
     destination_path:
       type: string
       description: >-
--- a/f/connectors/alerts/alerts_gcs.script.yaml
+++ b/f/connectors/alerts/alerts_gcs.script.yaml
@@ -29,6 +29,7 @@ schema:
       description: The name of the database table where alerts will be stored.
       default: null
       originalType: string
+      pattern: '^.{0,53}$'
     destination_path:
       type: string
       description: >-
--- a/f/connectors/comapeo/comapeo_observations.py
+++ b/f/connectors/comapeo/comapeo_observations.py
@@ -439,6 +439,7 @@ class CoMapeoDBWriter:
         comapeo_projects = outputs
 
         for project_name, project_data in comapeo_projects.items():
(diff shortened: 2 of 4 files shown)


f/connectors/alerts/alerts_gcs.script.yaml
```diff
@@ -29,6 +29,7 @@ schema:
       description: The name of the database table where alerts will be stored.
       default: null
       originalType: string
+      pattern: '^.{0,53}$'
```
IamJeffG, line 32:
```suggestion
      pattern: '^.{1,53}$'
```

f/connectors/comapeo/comapeo_observations.py
```diff
@@ -439,7 +439,9 @@ def handle_output(self, outputs):
         comapeo_projects = outputs
 
         for project_name, project_data in comapeo_projects.items():
-            existing_fields = self._inspect_schema(project_name)
+            # TODO: Safely truncate each project_name to 63 characters while retaining uniqueness
```
IamJeffG, line 442:
```suggestion
            # Safely truncate each project_name to 63 characters
            # TODO: ...while retaining uniqueness
```

f/connectors/kobotoolbox/kobotoolbox_responses.script.yaml
```diff
@@ -31,6 +31,7 @@ schema:
       description: The name of the database table where the form data will be stored.
       default: null
       originalType: string
+      pattern: '^.{0,54}$'
```
IamJeffG, line 34:
```suggestion
      pattern: '^.{1,54}$'
```

IamJeffG (approved):

Issue #22, opened 2024-12-10
[connectors] Update database records when upstream data source changes


rudokemper:
## Feature Request

Data from a source that has already been processed by a `connectors` script may change for various reasons, such as:

* A Kobo admin updates survey submissions on the KoboToolbox server for data QA purposes.
* A CoMapeo user edits an observation after it has been synchronized with the CoMapeo archive server and then synchronizes again.
* An alerts provider posts GeoJSON files to GCP, realizes these were added by mistake, and removes them. (They will persist in the data warehouse if already processed by the alerts script.)

Currently, our scripts do not update records in these scenarios. For instance, if I change the category of an already processed CoMapeo observation from "animal" to "fish," synchronize it with the CoMapeo archive server, and then run the CoMapeo script, the data warehouse will still reflect "animal" as the category.

Additionally, we should account for attachments. If media or files are deleted upstream, they should also be deleted from local storage. Likewise, if attachments are updated or replaced upstream, the corresponding local files should be updated as well.

I have confirmed this behavior with all three existing `connectors/` scripts (kobo, comapeo, alerts).

I can’t think of any use cases where we would want to preserve the original record or files after an update in the data source, but I’m open to thoughts. 

Barring any good reasons, let's adapt our scripts to consistently detect changes from the source, and if found, update database records and update attachments:

- [x] kobo
- [x] comapeo
- [x] alerts

PR #56, opened 2024-12-31
For connector scripts, update existing rows if changes detected

rudokemper:
## Goal

Refining the database writing process towards resolving https://github.com/ConservationMetrics/gc-scripts-hub/issues/22. For now, I've chosen to stick with KoboToolbox to maintain focus.

## What I changed

In looking at how we are doing DB writes, I conclude that the `safe_insert` method only inserts new rows and does nothing with existing ones. Consequently, `UniqueViolation` errors are avoided entirely, as no exceptions are raised for existing rows.

Hence, as a first attempt towards improving this, I am first checking to see if there is an existing row by running a `SELECT` query, and then if found, comparing the existing row's values with the incoming upstream data values to be inserted. If false, then insert the data as before, but I've added an `ON CONFLICT ... DO UPDATE SET` clause to handle rows that require an update.

As noted in #22, this is operating under the assumption that we will always want to update records if the upstream data source changes. So if any diff whatsoever is detected, then the row is updated.

Lastly, I find the logs in this KoboToolbox script to be rather sparse, so I added `inserted_count` and `updated_count` vars to track the number of inserts and updates, respectively, and logged those at the end.

While it seems to work, I am unsure if this is the most parsimonious or performative approach. I'm also mindful of the fact that @IamJeffG has petitioned several times that we should move to refactor these DB Writer classes as a whole, and so maybe this work is somewhat moot. Feedback welcome.

### DIFF
--- a/f/connectors/kobotoolbox/kobotoolbox_responses.py
+++ b/f/connectors/kobotoolbox/kobotoolbox_responses.py
@@ -443,7 +443,8 @@ class KoboDBWriter:
     @staticmethod
     def _safe_insert(cursor, table_name, columns, values):
         """
-        Safely construct and execute an INSERT query to avoid SQL injection.
+        Constructs and executes a safe INSERT query to prevent SQL injection,
+        and handles conflicts by updating existing records.
 
         Parameters
         ----------
@@ -458,17 +459,51 @@ class KoboDBWriter:
 
         Returns
         -------
-        None
+        tuple
+            A tuple containing the count of inserted and updated rows.
         """
+        inserted_count = 0
+        updated_count = 0
+
+        # Check if there is an existing row that is different from the new values
+        id_index = columns.index("_id")
+        values[id_index] = str(values[id_index])
+        select_query = sql.SQL("SELECT {fields} FROM {table} WHERE _id = %s").format(
+            fields=sql.SQL(", ").join(map(sql.Identifier, columns)),
+            table=sql.Identifier(table_name),
+        )
+        cursor.execute(select_query, (values[columns.index("_id")],))
+        existing_row = cursor.fetchone()
+
+        if existing_row and list(existing_row) == values:
+            # No changes, skip the update
+            return inserted_count, updated_count
+
         query = sql.SQL(
-            "INSERT INTO {table} ({fields}) VALUES ({placeholders})"
+            "INSERT INTO {table} ({fields}) VALUES ({placeholders}) "
+            "ON CONFLICT (_id) DO UPDATE SET {updates} "
+            "RETURNING (xmax = 0) AS inserted"
         ).format(
             table=sql.Identifier(table_name),
             fields=sql.SQL(", ").join(map(sql.Identifier, columns)),
             placeholders=sql.SQL(", ").join(sql.Placeholder() for _ in values),
+            updates=sql.SQL(", ").join(
+                sql.Composed(
+                    [sql.Identifier(col), sql.SQL(" = EXCLUDED."), sql.Identifier(col)]
+                )
+                for col in columns
+                if col != "_id"
+            ),
         )
 
         cursor.execute(query, values)
+        result = cursor.fetchone()
+        if result and result[0]:
+            inserted_count += 1
+        else:
+            updated_count += 1
+
+        return inserted_count, updated_count
 
     def handle_output(self, all_submissions):
         """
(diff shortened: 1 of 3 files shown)


f/connectors/kobotoolbox/kobotoolbox_responses.py
```diff
@@ -458,17 +459,51 @@ def _safe_insert(cursor, table_name, columns, values):
...
+
+        if existing_row and list(existing_row) == values:
+            # No changes, skip the update
+            return inserted_count, updated_count
+
         query = sql.SQL(
-            "INSERT INTO {table} ({fields}) VALUES ({placeholders})"
+            "INSERT INTO {table} ({fields}) VALUES ({placeholders}) "
+            "ON CONFLICT (_id) DO UPDATE SET {updates} "
+            "RETURNING (xmax = 0) AS inserted"
```
IamJeffG, line 485: `xmax = 0` is postgres magic.  Requires a code comment.

f/connectors/kobotoolbox/kobotoolbox_responses.py
```diff
@@ -458,17 +459,51 @@ def _safe_insert(cursor, table_name, columns, values):
...
+        select_query = sql.SQL("SELECT {fields} FROM {table} WHERE _id = %s").format(
+            fields=sql.SQL(", ").join(map(sql.Identifier, columns)),
+            table=sql.Identifier(table_name),
+        )
+        cursor.execute(select_query, (values[columns.index("_id")],))
+        existing_row = cursor.fetchone()
+
+        if existing_row and list(existing_row) == values:
+            # No changes, skip the update
+            return inserted_count, updated_count
```
IamJeffG, line 480: Why do we need everything above?  Won't the INSERT below just re-write the same values and the end result is the same?

IamJeffG: ah, is it so you can more accurately report on inserted_count, updated_count, without considering an exact overwrite to be an update?

IamJeffG (commented):
I am totally on-board using `INSERT ... ON CONFLICT UDPATE`.

I will say that this is some pretty magic looking code, hard to follow.  But probably correct and as well as we can do.  An idea to consider is to insert/update all the rows in a single INSERT statement, rather than a for loop, but I don't feel strongly about this.

Major question for you: Sometimes the primary key is a hash of the values.  If the values change, then so does the primary key: there is no conflict and so could we still end up with duplicate rows?

A different side of the same question is: Why do we think we want to update old rows at all?  Do we have some reason to trust today's values more than yesterday's values of yesterday's data?

f/connectors/kobotoolbox/kobotoolbox_responses.py
rudokemper: Yep, that's correct - otherwise everything that isn't inserted is counted as updated.

We absolutely do not _need_ it for the core ETL functionality of the script, but arguably, it's a useful feature for Windmill operators to understand what their runs are doing each time. 

It could be interesting to benchmark how the script performs with and without this code for the largest Kobo dataset that we are ingesting.

IamJeffG: I would document this, too, then.

Otherwise looks good to me!

rudokemper:
> Major question for you: Sometimes the primary key is a hash of the values. If the values change, then so does the primary key: there is no conflict and so could we still end up with duplicate rows?

Yeah, we'll have to carefully consider our approach to hashing UUIDs for the scripts that use them. KoboToolbox (and CoMapeo) provide their own unique identifiers so we don't have this problem here. 

For the alerts script: we hash alert records based on `alert_id`, and alert metadata based on `year`, `month`, `territory_id`, `type_alert`, and `description_alerts`. 

* I can reach out to the alerts provider to confirm that `alert_id` is static and will never change. 
* For metadata, maybe we omit `type_alert` and `description_alerts`.

I realize there are still potential pitfalls here (like the provider making a mistake), but I'm not sure what else we could do to avoid those.

> A different side of the same question is: Why do we think we want to update old rows at all? Do we have some reason to trust today's values more than yesterday's values of yesterday's data?

This is the million dollar question here. 

I think there are definitely some very good reasons why we want to update old rows. I've written some of them out in #22 for each connector script. These are real scenarios that have come up in practice.

That said, I have to say that it doesn't sit well with me to entirely erase the record of what was there before. 

Imagine a scenario where a new community organization administrator gets the keys to a KoboToolbox account. They review a previously created form, and decide to modify some of the submissions for data QA reasons in accordance to their best judgment. But in so doing, they erase some things that were recorded in a specific way with good reason. I have seen this happen many times at organizations, including ones that I have worked at. (And there is of course a more alarming version of this story where a bad actor gets the keys.)

Maybe in the future we can consider version control for data records. But in the absence of that, we have a binary choice between updating and not updating, and with that binary I definitely think we should update.

IamJeffG:
Thanks! Go ahead and merge at will.

PR #58, opened 2025-01-06
Document decision not to share code between scripts (for now)

rudokemper:
## Goal

We have discussed internally that for the time being, we will keep each script as self-contained as possible. Here I am explicitly documenting that in the repo's README.

I have also removed one TODO comment that is obviated by this decision.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,6 +32,9 @@ Note that Windmill also designates a `u/` directory for storing code per user on
 
 For information on developing scripts, see the [Windmill Scripts quickstart](https://www.windmill.dev/docs/getting_started/scripts_quickstart).
 
+While it is possible in Windmill to [share common logic across scripts with relative imports](https://www.
+windmill.dev/docs/advanced/sharing_common_logic) we have avoided doing so at this time. This decision aims to keep each script as self-contained as possible, promoting modularity and reducing interdependencies. As a trade-off, there may be some redundancy across common operations, such as writing to a database.
+
 You may develop within Windmill's code editor, or locally.  Developing locally has the advantage
 of being able to run tests.
 
--- a/README.md
+++ b/README.md
@@ -32,8 +32,7 @@ Note that Windmill also designates a `u/` directory for storing code per user on
 
 For information on developing scripts, see the [Windmill Scripts quickstart](https://www.windmill.dev/docs/getting_started/scripts_quickstart).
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -32,6 +32,9 @@ Note that Windmill also designates a `u/` directory for storing code per user on
 
 For information on developing scripts, see the [Windmill Scripts quickstart](https://www.windmill.dev/docs/getting_started/scripts_quickstart).
 
+While it is possible in Windmill to [share common logic across scripts with relative imports](https://www.
+windmill.dev/docs/advanced/sharing_common_logic) we have avoided doing so at this time. This decision aims to keep each script as self-contained as possible, promoting modularity and reducing interdependencies. As a trade-off, there may be some redundancy across common operations, such as writing to a database.
```
IamJeffG, line 36: On one hand I am totally open to re-visiting this in the future.  On the surface it seems silly we don't share common logic already.

On the other hand, I think one of the main reasons for struggle with what's documented at that link is that it ignores that each script creates its own (different) dependency stack.  So any shared code needs to work with a wide range dependencies and/or declare its dependencies.  Short of making it a proper package, there's not an easier way to do that.

This, I think, is worth documenting.

README.md
```diff
@@ -32,6 +32,9 @@ Note that Windmill also designates a `u/` directory for storing code per user on
 
 For information on developing scripts, see the [Windmill Scripts quickstart](https://www.windmill.dev/docs/getting_started/scripts_quickstart).
 
+While it is possible in Windmill to [share common logic across scripts with relative imports](https://www.
```
IamJeffG, line 35: avoid line break in the URL.  Markdown won't render it right.

rudokemper: Oops, a typo.

README.md
rudokemper: Something else I'm curious about, and couldn't find information about in the docs: does the Windmill UI include module files containing common code in the list of Scripts? That might be another reason to keep the status quo. I'd vote against having modules mixed in with the core scripts that our users are operating. (And something I'd want to document.)

IamJeffG: I think this is the answer, from that link:

> your scripts do not necessarily need to have a main function. If they don't, they are assumed to be shared logic and not runnable scripts.

rudokemper:
https://github.com/ConservationMetrics/gc-scripts-hub/pull/77 negates the decision taken in this PR.

PR #59, opened 2025-01-07
Add `alerts_twilio` script, and prepare `alerts_gcs` script for future usage in a Windmill Flow

### LINKED ISSUE #32: [alerts] Add an optional Twilio operation
## Feature Request

Currently, when new alerts are processed by the `alerts_gcs` script, either I or someone from the alerts provider sends a WhatsApp message to notify users about new alerts.

This approach has worked relatively well for our small user base, but there are some challenges:

* The process isn't deterministic; users often check in about alert statuses if they haven’t seen any by an expected date, perhaps assuming we’ve forgotten to notify them.
* There can be delays between the time alerts reach the data warehouse and when the notification is sent.
* If both the alerts provider and I are out of the office, no one is available to notify the users (though this hasn't happened yet).
* The users themselves have to also manually pass the message along to their constituencies, such as field rangers or community leaders.
* This process isn’t scalable; as we onboard more users and aim to grow to 20-30 in the coming years, managing this manually will become increasingly difficult.

We already anticipated this issue at the very beginning of the change detection project, and sought to solve the problem by adding automated messaging via Twilio in frizzle v1 (a private repo), which was envisioned to be triggered _only when_ all of the alerts have been processed (e.g. added to the warehouse and images stored in the datalake). 

See detailed technical notes [here](https://github.com/ConservationMetrics/frizzle-thespian/issues/118) and code [here](https://github.com/ConservationMetrics/frizzle-thespian/blob/main/components/twilio/__init__.py). _Note this is a private archive repo, if you need to see this content ask someone from CMI for access._ 

The Twilio code for frizzle v1 was relatively straightforward, and could possibly be rewritten for gc-scripts-hub. However, there is also a [`twilio` resource in Windmill](https://hub.windmill.dev/resource_types/123/twilio) so possibly a lot of work done already in this framework. **Let's first do some research to figure out the best approach.**

### User Stories & Acceptance Criteria

_to fill out more depending on research_

### Technical notes

For `send_twilio_message`, we will need the following config vars

* `account_sid` (String)
* `auth_token` (String)
* `message_service_sid` (String)
* `template_sid` (String, optional)
* `origin_number` (String)
* `forward_numbers` (Array)

`send_twilio_message` should send only one message. For the message template, let's start with the following. These values should be pulled from `alerts__metadata`.

> **`{total_alerts}`** new change detection alert(s) have been published on your alerts dashboard for the month of `{month}`. The following activities have been detected in your region: **`{description_alerts}`**. Visit your alerts dashboard here: `{dashboard_link}`

* For `month`, we can use a lookup table to rewrite the value as an actual month e.g. 4 = April. 
* For `description_alerts`, let's rewrite the value so that `gold_mining,logging,oil_drilling` becomes `gold mining, logging, oil drilling`.

rudokemper:
## Goal

Closes #32.

The vision here is that these two scripts will be chained together with the CoMapeo alerts script in a Windmill Flow via a future PR:

1. `alerts_gcs` runs. If new rows are written to the alerts or metadata tables, then the script returns an `alerts_statistics` dictionary; if not, it will return None.
2. `comapeo_alerts` runs. Dependent on https://github.com/ConservationMetrics/gc-scripts-hub/pull/60 being merged.
3. `alerts_twilio` runs. I expect that in the Flow I can configure this not to run if `alert_gcs` does not return anything.

## Screenshots

_A WhatsApp message I received from our Twilio service bot:_

image

## What I changed

**`twilio_message_template`**

* I added a `twilio_message_template` resource type. 
* I opted to create my own rather than use the [verified `twilio` resource in windmill](https://hub.windmill.dev/resource_types/123/twilio) because this resource only provides `auth_token` and `account_sid` arguments, whereas we also need `content_sid` and could provide an additional `message_service_sid` for metrics tracking. 
  * Additionally, I sought fit to add `origin_number` and `recipients` so as to bundle all of the relevant Twilio config parameters in one resource: to me, it seems more logical to configure all of the Twilio config (including phone numbers that will be receiving the message) in one screen, rather than to parse those between a resource and the script itself.

**`alerts_gcs`**

* The script now compiles `alert_statistics` in `prepare_alerts_metadata()`, consisting of values that are inserted in the Twilio message. 
  * Here, we face a conundrum of needing to choose a row to generate alerts statistics for, and need to make an assumption that the alerts provider is always posting new alerts for the last month only. Grounded in that assumption, I am grouping and sorting the alerts metadata Dataframe in descending order by year and month, so as to always have the latest row at `iloc[0]`.
* The DB Writer now returns a new boolean `new_alerts_data` which is set to true if any alerts rows or metadata rows were written. 
* If `new_alerts_data` is true, then the script as a whole returns `alerts_statistics`, with the expectation that the `alerts_twilio` script will be able to use this as the next step in a Flow.

**`alerts_twilio`**

* This new script takes three properties: 
  * an `alerts_statistics` object
  * a `twilio` object referring to the above resource type
  * a `territory_name` string that represents the slug used for subdomains. For example, `demo` for the demo, and `m**a`, `t**p`, ...
* The script calls a `send_twilio_message` function which leverages the Twilio client library to send a message to the API with the expected variables.

**Tests**

* I adapted the CSV fixture to include a row with the latest month and year somewhere in the middle, and added a test to assert that `alerts_statistics` nevertheless returns values from this new row (to test the Pandas Dataframe sorting behavior).
* The `alerts_gcs` e2e test now checks to see the script returns an `alerts_statistics` the first time, and None the second time (since this second run will update 0 alerts and alerts__metadata rows, so the script should not return anything).

## What I'm not doing here

* Mocking or testing the Twilio API, since our usage of the API is entirely orthodox.
* Setting up the Flow itself that will chain together `alerts_gcs`, the CoMapeo POST alerts script to be merged from https://github.com/ConservationMetrics/gc-scripts-hub/pull/60, and `alerts_twilio`. When I do that, I expect that I will make some changes to what `alerts_gcs` actually returns in order to trigger runs for the other two scripts.


f/connectors/alerts/alerts_gcs.script.yaml
```diff
@@ -53,6 +55,21 @@ schema:
         will be retrieved. This ID is used to filter and process relevant
         alerts.
       default: null
+    territory_name:
```
rudokemper, line 58: I'm not convinced on this variable name, as it is ambiguous. 

A community operator might misunderstand this to refer to the real name of their territory, not the GC subdomain slug e.g. `{slug}.guardianconnector.net`.

On the other hand, "slug" is not a commonly known term for non-technical users. 

Maybe "subdomain" is a better choice.

rudokemper:
Being in the weeds of implementation, I missed something crucial here: that a message should be sent _if and only if_ new alerts were written to the db. Otherwise, a Twilio message will be circulated every time the script is run.

Additionally, I would like to reconsider the sequence of when a Twilio message is actually sent, also taking into consideration that alerts will be posted to the CoMapeo API.

I'll re-open when this is ready again.

f/connectors/alerts/README.md
```diff
@@ -1,8 +1,8 @@
-# Google Cloud Alerts Change Detection Integration
+# `alerts_gcs`: Google Cloud Alerts Change Detection Integration
 
-This script fetches change detection alerts and images from a storage bucket on Google Cloud Platform. The script transforms the data for SQL compatibility and stores it in a PostgreSQL database. Additionally, it saves before-and-after images -- as TIF and JPEG -- to a specified directory.
+This script fetches change detection alerts and images from a storage bucket on Google Cloud Platform. The script transforms the data for SQL compatibility and stores it in a PostgreSQL database. Additionally, it saves before-and-after images -- as TIF and JPEG -- to a specified directory.
```
IamJeffG, line 3:
```suggestion
This script fetches change detection alerts and images from a storage bucket on Google Cloud Platform. The script transforms the data for SQL compatibility and stores it in a PostgreSQL database. Additionally, it saves before-and-after images -- as TIF and JPEG -- to a specified directory.
```

removes trailing space

f/connectors/alerts/README.md
```diff
@@ -1,8 +1,8 @@
-# Google Cloud Alerts Change Detection Integration
+# `alerts_gcs`: Google Cloud Alerts Change Detection Integration
```
IamJeffG, line 1: I might prefix all of this with a sentence explaining there are 2 scripts in this folder and described in this README:

```suggestion
This README introduces two scripts that live in this folder:

- `alerts_gcs`: Google Cloud Alerts Change Detection Integration
- `alerts_twilio`: Send a Twilio Message


# `alerts_gcs`: Google Cloud Alerts Change Detection Integration
```

f/connectors/alerts/README.md
```diff
@@ -15,31 +15,35 @@ Google Cloud storage has a built-in API solution, [GCS JSON API](https://cloud.g
...
 **Raster:**
 Currently, we are assuming there to be only four raster images for each change detection alert: a 'before' and 'after' used for detection and visualization, respectively.  Each of these is saved in both TIFF and JPEG format in the following way:
 
-```
-<territory_id>/<year_detec>/<month_detec>/<alert_id>/images/<sat_viz_prefix>_T0_<id>.tif
-<territory_id>/<year_detec>/<month_detec>/<alert_id>/images/<sat_viz_prefix>_T0_<id>.jpg
-...
+    <territory_id>/<year_detec>/<month_detec>/<alert_id>/images/<sat_viz_prefix>_T0_<id>.tif
+    <territory_id>/<year_detec>/<month_detec>/<alert_id>/images/<sat_viz_prefix>_T0_<id>.jpg
+    ...
```
IamJeffG, line 41: do you want to talk about the return value?


```suggestion
    ...
    
### Returns

If new rows are written to the alerts or metadata tables, then the script returns an `alerts_statistics` dictionary; otherwise it will return `None`.
```


or fine to hold off til a follow-up PR, since I saw you expect the details of this return value may need to change

f/connectors/alerts/alerts_gcs.py
```diff
@@ -321,17 +336,33 @@ def prepare_alerts_metadata(alerts_metadata, territory_id):
...
     # Filter DataFrame based on territory_id
     filtered_df = df.loc[df["territory_id"] == territory_id]
 
+    # Group the DataFrame by month and year, and get the last row for each group
+    filtered_df = (
+        filtered_df.loc[df["territory_id"] == territory_id]
+        .groupby(["month", "year"])
+        .last()
+        .reset_index()
+    )
```
IamJeffG, line 359: I think you might avoid the `.reset_index()` by instead passing `as_index=False` in the groupby

https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.groupby.html




```suggestion
    filtered_df = (
        filtered_df.loc[df["territory_id"] == territory_id]
        .groupby(["month", "year"], as_index=False)
        .last()
    )
```

f/connectors/alerts/alerts_gcs.py
```diff
@@ -321,17 +336,33 @@ def prepare_alerts_metadata(alerts_metadata, territory_id):
...
+
+    # Sort the DataFrame by year and month in descending order
+    filtered_df = filtered_df.sort_values(
+        by=["year", "month"], ascending=[False, False]
+    )
```
IamJeffG, line 364: nit: I'd probably move this call to be in the same line as your `.iloc[0]` below.  That line **depends on** this having been run, and you do not want to risk someone changing one of the lines without the other.

Also I think nothing else in this file depends on this sorting besides the `.iloc[0]`, so you don't even necessarily need to persist the result of the sort.

IamJeffG (approved):
I agree about the new `twillio_message_template` resource type.

f/connectors/alerts/alerts_gcs.py
rudokemper: Thanks, good catch!

PR #62, opened 2025-01-15
Add connector script for Locus Map points data

### LINKED ISSUE #61: [locusmap] Add script to extract and transform uploaded Locus Map data
## Feature Request

Add a script that takes a temp path for Locus Map data (CSV/KML/GPX or ZIP with media attachments), transforms the data to write to a database in the same format as CoMapeo / Kobo data (geometry fields prefixed with `g__`), and copies attachments if any to the datalake.

The process by which the Locus Map data is manually uploaded to volume mount storage will be handled in a different issue.

rudokemper:
## Goal

Closes #61. Adds support for writing Locus Map point exports in CSV, KML, GPX formats, as well as KMZ and ZIP which include media attachments.

## What I changed

* Added a `locusmap.py` script to handle ETL for a LocusMap export file in the above formats.
  * For KML and GPX, being both XML-format files, I used `etree` from `lxml` to parse the files. Initially, I tried to use [`gpxpy`](https://pypi.org/project/gpxpy/) for GPX, but I found that it does not support parsing multiple `<link>` attributes per waypoint.
  * Instead of assuming that the incoming Locus Map is a temp file as before, I adopted @IamJeffG's suggestion to have a "Delete Locusmap Export File?" boolean and utilize that to handle optional deletion.
* Added real assets exported from Locus Map, and tests for each format checking for relevant transformations and attachment handling. I also added one test that checks that the script will throw a ValueError for an unsupported format.

## What I'm not doing here

* Dealing with track (LineString) exports.


f/connectors/locusmap/README.md
```diff
@@ -0,0 +1,38 @@
+# Locus Map: Import Data
+
+This script imports data from a Locus Map export file into a database table. It reads a file containing spatial data, transforms the data into a structured format, and inserts it into a PostgreSQL database table. Additionally, it downloads any attachments associated with the spatial data and saves them to a specified directory. It then deletes the temporary files.
+
+Locus Map exports data as a CSV, KML, and GPX. If attachments are included, then the export will be as a ZIP file compressing the spatial data file together with a directory containing the attachment files. (It is also possible to export data as a DXF or Ov2 file, but these are not commonly used formats, so this script does not intend to support them.)
+
+
+> [!IMPORTANT]
+> This script assumes that the temporary CSV file and attachments directory should always be deleted after processing.
```
IamJeffG, line 9: I'm fine with this, but just another idea:

You could consider putting a checkbox "Delete after ingest?" on the Windmill form. And I'd claim that if you did that, then you would no longer have to refer to this as "temporary" i.e. no longer need to make assumptions about where this file lives or what is the user's plan with it.

That said ... it would probably never get used this way :-)

f/connectors/locusmap/README.md
```diff
@@ -0,0 +1,38 @@
...
+
+[^1]: The web app could use something like the [Uppy](https://github.com/transloadit/uppy) JavaScript Library to upload files, and call `uppyDashboard.on('complete', (result) => {}` to POST a Windmill API request after successful file upload.
+
+
+## Sample curl request for Windmill API
+
+```
+curl -X POST \
+  https://windmill.demo.guardianconnector.net/api/w/frizzle-demo/jobs/run/p/f/connectors/locusmap/locusmap \
+  -H 'Authorization: Bearer Your_Access_Token' \
```
IamJeffG, line 24: maybe mention where's this access token come from

IamJeffG (commented):
so far so good!

f/connectors/locusmap/README.md
rudokemper: I like this idea! Thanks for the suggestion.

IamJeffG (approved):

PR #69, opened 2025-01-24
Improve the CoMapeo alert script for runtime Windmill usage

rudokemper:
## Goal

Now that the CoMapeo Cloud server has the `GET projects/:projectId/remoteDetectionAlerts` endpoint that we needed, and I could deploy a new image for testing, I found that some improvements were needed for Windmill runtime usage of the CoMapeo alerts script.

## Screenshots

image

## What I changed

* Added missing Python requirements added to `comapeo_alerts.script.lock`
* The CoMapeo server expects remote detection alerts in a typed schema, so I added a function to transform alerts to be consistent with that schema. And adapted the tests to reflect this new behavior.
* Minor improvements to logging for clarity, and a few type annotation corrections.

### DIFF
--- a/f/connectors/comapeo/comapeo_alerts.py
+++ b/f/connectors/comapeo/comapeo_alerts.py
@@ -52,6 +52,10 @@ def main(
 
     unposted_alerts = filter_alerts(comapeo_alerts_endpoint, comapeo_headers, alerts)
 
+    if not unposted_alerts:
+        logger.info("No new alerts to post!")
+        return
+
     transformed_unposted_alerts = transform_alerts(unposted_alerts)
 
     post_alerts(comapeo_alerts_endpoint, comapeo_headers, transformed_unposted_alerts)
@@ -83,6 +87,9 @@ def get_alerts_from_db(db_connection_string, db_table_name: str):
     ]
     cur.close()
     conn.close()
+
+    logger.info(f"{len(alerts)} alerts found in database.")
+
     return alerts
 
 
@@ -112,6 +119,7 @@ def _get_alerts_from_comapeo(comapeo_alerts_endpoint: str, comapeo_headers: dict
 
     posted_alert_source_ids = {alert["sourceId"] for alert in alerts}
 
+    logger.info(f"{len(posted_alert_source_ids)} alerts found on CoMapeo.")
     return posted_alert_source_ids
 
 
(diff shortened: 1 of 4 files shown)


f/connectors/comapeo/comapeo_alerts.script.lock
IamJeffG, line 1: I'm not certain but my impression is that Windmill will overwrite this file after changes to the `.py` script.  Windmill generates (or at least, _can generate_) this file automatically from parsing the `# requirements:` comment block at the top of the `.py` script.

All this to say: Should these new requirements be added there?

f/connectors/comapeo/comapeo_alerts.py
```diff
@@ -144,13 +156,56 @@ def filter_alerts(
...
+    """
+    logger.info("Transforming alerts...")
+
+    transformed_alerts = [
+        {
+            # CoMapeo API requires these to be ISO 8601 format
+            "detectionDateStart": datetime.strptime(
+                alert["date_start_t0"], "%Y-%m-%d"
+            ).isoformat(timespec="seconds")
+            + "Z",
```
IamJeffG, line 185: `"%Y-%m-%d"` is already a valid ISO-8601 _date_, but maybe CoMapeo's requirement is that the value be a _datetime_.


Just want to double-check that the addition of time here is happening as you expect:
- is midnight an acceptable time? In particular for `detectionDateEnded`, if the end was, say noon on that date, then your new timestamp would be before the actual end.
- is UTC correct? Do we have any way to infer a more correct timezone?


Just ideas -- I trust you've thought about this already.

rudokemper: Thanks, noted the datetime requirement in the comment.

Yeah, this is quite tricky. There is no deterministic way to infer. But your notes made me want to adapt the timestamp to 15:00 UTC which is an expected time for when the alerts provider might post alerts on the day.

f/connectors/comapeo/comapeo_alerts.script.lock
rudokemper: I think I could benefit from improving my understanding of dependency management for scripts in Windmill. Without me adding these requirements, the script failed in the Windmill UI. I'll ask you about this later on a call, probably easier to quickly talk though.

Issue #65, opened 2025-01-17
[locusmap] Prototype Windmill App to upload a Locus Map export file and trigger script


rudokemper:
## Feature Request

Let's create a Windmill App which takes a Locus Map upload file (CSV, GPX, KML, KMZ, or ZIP), and adds an event listener to trigger the `locusmap.py` script to run (once https://github.com/ConservationMetrics/gc-scripts-hub/issues/61 is merged).

In the future, we will want a more sophisticated Uploader app that can deal with files from many different sources aside from Locus Map, but we can stay focused on Locus Map for now to have a working proof-of-concept for the "GC Uploader" vision.

PR #72, opened 2025-02-03
Add a "GC Uploader" Windmill app that takes a Locus Map export and triggers the connector script to run

rudokemper:
## Goal

Closes https://github.com/ConservationMetrics/gc-scripts-hub/issues/65 by adding a "GC Uploader" app that is designed to take a Locus Map export file (CSV/KML/GPX or ZIP with attachments), and triggers the Locus Map connector script after successful upload.

## Screenshots

image

## What I changed

* I added an app with three reactive components and state management:
  * `selectFile` allow the user to select a Locus Map export file
  * `uploadFile` (activated once the user selects a file), when pressed, will trigger the Locus Map connector script to run
  * `resultMessage` will show a status message of "Processing...", "File uploaded successfully!", or nothing depending on app state.
* To accomplish all of this, there are a few inline scripts that set reactive state variables, triggered by each component.
* There is also an inline script "Save data to disk" to handle the file upload, which could later be turned into a standalone Windmill script.

## What I'm not doing here

* Setting this up to allow users to upload any data without specifying the type, ala Felt's "Upload Anything" button. For now, as an initial working proof-of-concept, this app is only set up to work with Locus Map data.
  * Following the merging of this PR, I think I would like to do some schematizing of how to move forward from here to support other kinds of uploads.
* Allowing the user to change the name of their database table; for now, it is taking the filename point blank.
* Styling and design of the app. The look-and-feel is out of the box from Windmill apps.
* The "Save data to disk" script to handle the file upload was added to the app as an inline script. We may want to make it a standalone script so it can be used by other Windmill apps.
* Error handling. If something goes wrong in the Locus Map connector script, I am not updating `resultMessage` (but the Windmill app UI does show an error message modal out-of-the-box). The reason I did not do so is because I would need to modify the Locus Map script itself to return that error message, and I didn't want to mess with the script itself for now.
* I have not yet added documentation. I will do so after I receive feedback on the approach to the app.

## Notes to the reviewer

* I find that the app YAML, which was 100% generated by the Windmill app builder UI, is pretty unreadable! I think the app is best reviewed by syncing the code to Windmill and checking it out in the app builder interface.  (CMI's `demo` Windmill instance already has the app synchronized to it). You can use one of the test assets from the Locus Map connector script.
* I am more than happy to give a demo / walkthrough of the code! I think that is likely the easiest way to understand how all of this works.


IamJeffG (approved):

rudokemper:
I added documentation about apps in general, and this specific app. I also scrubbed the app from my demo config values (like the postgres resource) and described how to configure the app for usage.

@IamJeffG maybe you want to have a quick read-through of the README changes.

IamJeffG (approved):
LGTM. Did you have any specific questions?

rudokemper:
None other than if it reads clearly to you. So all good here, thanks.

PR #74, opened 2025-02-04
Remove "frizzle" prefix from persistent storage paths

rudokemper:
## Goal

In tandem with https://github.com/ConservationMetrics/gc-forge/pull/47 and responding to the [edge case](https://github.com/ConservationMetrics/gc-forge/issues/43#issuecomment-2632536556) that emerges from making that change, the goal here is to remove reference to frizzle in persistent storage paths. Then we can be done with this issue once in for all in our codebases.

I'm sufficiently motivated to make the change in our handful of community deployments too, for the sake of consistency.

### DIFF
--- a/f/connectors/alerts/alerts_gcs.py
+++ b/f/connectors/alerts/alerts_gcs.py
@@ -44,7 +44,7 @@ def main(
     territory_id: int,
     db: postgresql,
     db_table_name: str,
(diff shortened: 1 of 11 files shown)


IamJeffG (approved):
I think you can rename the container-side of these volume mounts in CapRover, without having to modify where the volume gets mounted on the host VM

Issue #31, opened 2024-12-11
Add ODK script


rudokemper:
In a past `frizzle` repo, we created a data flow (script) to download and store ODK data from the v1 API (returned as XML).

I don't think this was the right approach. Better options:

*  The [OpenRosa API](https://docs.getodk.org/central-api-openrosa-endpoints/) is the preferred API which returns more of the survey response data in JSON.
* Better yet, we could use [pyODK](https://getodk.github.io/pyodk/) (providing that it returns all of the data that we want).

Here is some past discussion of what's been an issue with the v1 API:

> The KoboToolbox (v1) API returns a special `_geolocation` field for submissions, which inherits any geolocation data that was captured during form entry. This enables us to store Kobo submissions in a GeoJSON-esque format, with the expected geometry fields of type and coordinates.
> 
> However, the ODK API does not return any special geolocation fields for submissions, which means that any data pulled by the ODK component currently is being stored with null coordinates. Any geolocation fields which do store coordinates are treated just like normal fields, so we'd have to either know in advance which fields are being used to capture coordinates, or iterate through the values to find anything that resembles a valid lat/long pair and pass that to `geometry/coordinates`.

Additionally, for reference, the past work to fetch and enqueue ODK data from the v1 API in the [thespian Python framework](https://thespianpy.com/doc/)::

```python
class ODKPull(TimerMixin, Component):

    def __init__(self, name="odk-pull", initial_config={}):
        super().__init__(name, initial_config)

    def fetch_submissions(self):
        self.log.debug(f"Fetching submissions now!")

        base_uri = self.local_config.get("uri")

        # The ODK API requires a user to first authenticate and get a session token.
        credentials = {
            "email": f'{self.local_config.get("account")}',
            "password": f'{self.local_config.get("pass")}',
        }
        session_data = json.dumps(credentials)
        session_headers = {"Content-Type": "application/json"}
        session_uri = f"{base_uri}/v1/sessions"

        session_response = requests.post(
            session_uri, data=session_data, headers=session_headers
        )

        token = json.loads(session_response.text).get("token")

        # Now that we've authenticated and started a session with the API,
        # We can provide this token to the request headers.
        headers = {
            "Authorization": f"Bearer {token}",
        }

        for form_id in self.local_config.get("forms"):
            project_id = self.local_config.get("project", {})
            seen_submissions = self.local_config.get("seen_submissions", {})
[...]

PR #76, opened 2025-02-12
Add ODK connector script

rudokemper:
## Goal

Closes https://github.com/ConservationMetrics/gc-scripts-hub/issues/31.

## What I changed

* Added a `c_odk_config` resource type that supplies the variables required by pyODK. (I will try to submit this resource type to the Windmill hub too, and if approved, remove this file and the `c_` prefix in the script.)
* Added an `odk_responses` script that utilizes the pyODK client library to handle requests to an ODK Central server. This really streamlines the process compared to previously explored approaches (in `frizzle-thespian`) to make calls to the API directly.
  * Otherwise, this script is very much similar to the kobo one.
* To test the script, I've opted to mock the server responses for requests that pyODK makes under the hood.
* For the testing env, I am storing pyODK cache in a temp dir so that the (mocked) token created by pyODK doesn't get stored in local cache. I am also squelching a warning about an upstream pyODK issue about a deprecated usage issue.

## What I'm not doing here

* The quantity of duplicated code between scripts is becoming out of hand... so I've got a forthcoming PR stacked on this one to solve that problem.


f/connectors/odk/tests/assets/server_responses.py
IamJeffG, line 1: It's ok if you want to leave this as-is, but IMHO it's overkill to make these mocked JSON responses into a Python module.

You may have seen this pattern elsewhere and the reason we did it there was because the response needed to be parameterized -- and then things got a bit out of hand with copy-paste.

I haven't tested it with windmill syncing in particular, but other than that, you should be able to store each of these 3 things in their own `asset_foo.json` file, and have the `conftest.py` fixture read that file from disk.

IamJeffG (commented):
I did not really look at the `odk_responses.py` because I am sure it's quite similar to our other code and you know it best.  LMK if there's something specific you want my attention on there, however.

f/connectors/odk/tests/assets/server_responses.py
rudokemper: Ok. I saw the pattern in our other tests for similar fetch scripts and assumed there was a good reason behind it, so didn't think to question it.

PR #77, opened 2025-02-12
Implement shared code to be used across scripts

rudokemper:
## Goal

To move towards DRY code by introducing files containing shared logic between the different scripts. 

## Screenshots

This is what a Windmill workspace looks like with this change introduced:

image

I wonder if we can prevent anything in `f/common_logic` from showing up on the Windmill Operator view. Not sure if that would negatively impact the ability of other scripts to import from these files, though.

## What I changed

* Created `f/common_logic/db_connection.py` and `f/common_logic/db_transformations.py` to store code relating to making a postgres connection, and transforming rows, respectively.
* Imported code from these files in all of the scripts where applicable.

## What I'm not doing here

* Refactoring DB Writer classes, or move around any other code where a refactor is required to achieve an abstraction. For now, I've just copied over functions that were repeated verbatim across the scripts, like `conninfo()`. The intent here is to lay the foundation for more shared code usage in the future.


rudokemper:
If we are agreed on this approach, then I will close https://github.com/ConservationMetrics/gc-scripts-hub/pull/58.

IamJeffG:
> I wonder if we can prevent anything in f/common_logic from showing up on the Windmill Operator view. Not sure if that would negatively impact the ability of other scripts to import from these files, though.

The only place I can think of looking is in group permissions on the folder.  Specifically in `f/common_logic/folder.meta.yaml` (create it if necessary), you can declare who (both groups and specific users) can do certain things to that folder.  Example of one in a different deployment:

```yaml
summary: null
display_name: frizzle
extra_perms:
  g/all: false
  u/jeffg: true
owners:
  - u/jeffg
```

We'd have to dig into documentation for this file, or maybe you can set the folder to be not viewable in the Windmill webapp directly (and then see what folder.meta.yml it generated by `wmill sync pull`.  

I strongly doubt (?) that making these not visible to certain users would prevent other scripts from importing them.

IamJeffG (approved):

rudokemper:
I really wanted your suggestion to be true, but unfortunately it doesn't work. You can hide the `common_logic` folder from a user via permissions, but then their script runs with module imports run into errors:

```
TypeError: 'module' object is not callable
  File "/tmp/windmill/wk-default-48edfaa1c14a-uP7rX/01950263-6a8b-33f0-ac04-4bc8c3463d50/f/connectors/comapeo/comapeo_observations.py", line 62, in main
    db_writer = CoMapeoDBWriter(conninfo(db))
                                ^^^^^^^^^^^^
```

Granting permission for the user to access the folder causes their runs to successfully import and work again.

There is something you can do in the UI to merely filter out scripts that don't have a `main` function from the view, but I've yet to find a way to change the default (which is to show them). That would be an acceptable workaround:

image

Barring that, I guess the only implication for this PR and future work is to consider how many module files to keep. We don't want them to clutter the actual scripts in the UI. 

But then again, we can always train folks to click into specific folders like `f/connectors` or `f/export` as a first step. It's not that big of a deal.

rudokemper:
I will merge this for now as the pros outweigh the cons, and we can figure out optimal workarounds in the future.

rudokemper:
Postscript: I was wrong! In Windmill, Operator users already don't see any scripts without a main function in their Workspace. But the scripts that import from them work all the same. So this was a non-issue as it's fine to expect Developers or Admins to know to ignore (or filter away) module scripts, and we can spare our Operator users the confusion of what these are.

IamJeffG:
Great solution!

Issue #30, opened 2024-12-11
Add script to download Survey123 responses


rudokemper:
It seems quite feasible that we can add a data flow for consuming submissions data from an ArcGIS Survey123 form. This is a nice feature to add to `connectors`; with a Survey123 data flow added, we can say that GuardianConnector supports data integration from most of the major data collection apps being used by communities today, including the popular enterprise one.

There are two ways of doing this.

## ArcGIS REST API

To do this, we can leverage the [ArcGIS API query operation](https://developers.arcgis.com/rest/services-reference/enterprise/query-feature-service-layer/) on the feature service layer storing all of the Survey123 results, like this: `https://{subdomain}.arcgis.com/{service_id}/arcgis/rest/services/{feature_id}/FeatureServer/query`

A simple way to get all of the data is using the following params:
```python
    params = {
        'where': '1=1',  # This will query all records
        'outFields': '*',  # This will fetch all fields
        'f': 'geojson',  # Request the response in GeoJSON format
        'returnGeometry': 'true'  # Include geometry in the response
    }
```

Then, to get the attachments, we can iterate through all of the returned features' `object_id` and make the following request `https://{subdomain}.arcgis.com/{service_id}/arcgis/rest/services/{feature_id}/FeatureServer/{object_id}/attachments`. The response will be a GeoJSON file with URLs to download for all attachments for that `object_id`.

This will work for any Survey123 feature service layers that are public; for anything that is private, we will need to [generate a token](https://developers.arcgis.com/rest/services-reference/enterprise/generate-token/) in exchange for user credentials, like this:

```
curl -X POST "https://www.arcgis.com/sharing/rest/generateToken" \
     -d "username=your_username" \
     -d "password=your_password" \
     -d "client=requestip" \
     -d "f=json"
```

This will return a token. You would then append `&token=` to the request params.

There is a script available in our (private) [`gc-programs`](https://github.com/ConservationMetrics/gc-programs/tree/main/scripts/download-arcgis-data) repo that leverages the ArcGIS REST API and could be retooled for Windmill.

## ArcGIS API for Python

We could also look at using the [ArcGIS API for Python](https://developers.arcgis.com/python/latest/guide/get-started/). This requires you to have an ArcGIS account. But the process could be something like the following:

```python
from arcgis.gis import GIS
from arcgis.features import FeatureLayer

# Authenticate (replace with your credentials or use `gis = GIS()` for interactive login)
[...]

PR #81, opened 2025-03-04
Add two `connector` scripts: ArcGIS Feature Layer (for Survey123 or other), and GeoJSON to Postgres

rudokemper:
## Goal

Closes https://github.com/ConservationMetrics/gc-scripts-hub/issues/30.

This PR introduces:

1. A script to download the contents of an ArcGIS Feature Layer, including attachments, using the ArcGIS REST API. (Survey123 results are stored as a Feature Layer, with the same schema as a generic Feature Layer.)
2. A script to write a GeoJSON file to Postgres.

#### Why both scripts?

ArcGIS Feature Layer content can be requested directly as GeoJSON, so rather than building a specialized ingestion script, I created a reusable **"write GeoJSON to the data warehouse"** script. The ArcGIS script leverages this to handle database writes.

## What I changed

* Created an `c_arcgis_account` resource type composed of username and password (which is what is required by the ArcGIS  API to exchange for a token to make requests).
* Created an `arcgis_feature_layer.py` script that requests an access token, feature layer content, and attachments from the ArcGIS API. The structure of this script is broadly similar to those of the other connector scripts, specific leverage of the API endpoints aside.
* Created a `geojson_to_postgres.py` script that takes a GeoJSON file path, applies "`geometry()` --> `g__type` & `g__coordinates`" transformations, and writes to database, with an optional "Delete GeojSON?" boolean and function to delete the file from disk if true. (Similar to the Locus Map script.)
* e2e tests for both scripts. These are also in-line with how we do tests for other connector scripts. For the ArcGIS script, server responses are provided that match the schema of what the ArcGIS REST API returns. (I tested this with a real ArcGIS account.)

## Future extensions

(1) Much of the data we ingest (e.g., from CoMapeo or Locus Map) follows a GeoJSON-esque schema anyway—where `g__type` and `g__coordinates` store the spatial data that would be in a feature's `geometry` object, and the rest of the columns are its properties. A similar approach could work for those scripts:

* Download attachments
* Apply necessary transformations for GeoJSON compliance
* Use `geojson_to_postgres` to handle database writes

This would minimize redundant DB write logic and, if we ever adopt PostGIS, allows us to update the implementation in one place.
 
(2) Additionally, we might see fit to rewrite the `gc_uploader` app to accept either a GeoJSON file, a Locus Map export, or other formats in the future. The app would detect the file's data schema and apply any necessary transformations to ensure GeoJSON compliance before passing it to the `geojson_to_postgres` script. The broader goal here is to move toward an "Upload Anything" UI—similar to Felt—where users can upload different geospatial data formats seamlessly.

If we agree on the above two possible extensions, I will create issues to track the proposed refactors.


f/connectors/arcgis/README.md
```diff
@@ -0,0 +1,13 @@
...
+
+Usage of this script requires you to have an ArcGIS account, in order to generate a token.
+
+The feature layer URL can be found on the item details page of your layer on ArcGIS Online:
+
+![Screenshot of a feature layer item page](arcgis.jpg)
+
+This script uses the [ArcGIS REST API Query Feature Service / Layer](https://developers.arcgis.com/rest/services-reference/enterprise/query-feature-service-layer/) endpoint.
+
+Note: we have opted not to use the [ArcGIS API for Python](https://developers.arcgis.com/python/latest/) library because it requires installing `libkrb5-dev` as a system-level dependency. Workers in Windmill can [preinstall binaries](https://www.windmill.dev/docs/advanced/preinstall_binaries), but it requires modifying the Windmill `docker-compose.yml`, which is too heavy-handed an approach for this simple fetch script.
```
IamJeffG, line 13: :+1:

f/connectors/geojson/README.md
```diff
@@ -0,0 +1,3 @@
+# `geojson_to_postgres`: Upload a GeoJSON file to the data warehouse
+
+This script imports data from a GeoJSON file into a database table. It reads a file containing spatial data, transforms the data into a structured format, and inserts it into a PostgreSQL database table. Optionally, it then delete the export file.
```
IamJeffG, line 3: I would highlight front-and-center some unique characteristics about the Geojson -> pg 

1. It flattens properties into their own columns
2. It's not using PostGIS


This is far from the only (or even most intuitive?) way to store geojson in the database.  For example other ways  might be:
1. just create a single JSONB column and throw the entire blob in there.
2. Install PostGIS extension, and then you can
    - represent the "geometry" more faithfully.
    - just use ogr2ogr to inject, or I wouldn't be surprised if PostGIS has other functions to help flatten json.

I'm not suggesting you switch to either of these -- just make clear what it is that we are doing.

IamJeffG: Also document that it looks for a uniique "id" as primary key.  And where it that, inside "properties" or outside?

f/connectors/arcgis/tests/conftest.py
```diff
@@ -0,0 +1,77 @@
...
+
+
+@pytest.fixture
+def mocked_responses():
+    with responses.RequestsMock() as rsps:
+        yield rsps
+
+
+@pytest.fixture
+def arcgis_server(mocked_responses):
```
IamJeffG, line 18: nice

f/connectors/arcgis/arcgis_feature_layer.py
```diff
@@ -0,0 +1,274 @@
...
+
+import json
+import logging
+import os
+from pathlib import Path
+
+import requests
+
+from f.common_logic.db_connection import postgresql
+from f.connectors.geojson.geojson_to_postgres import main as save_geojson_to_postgres
```
IamJeffG, line 13: You asked about calling another script from this script. I see no problem with it. Let's try it out and see how it goes.

PR #83, opened 2025-04-01
Include `confidence` field for hashing alerts_metadata UUID

rudokemper:
When working on https://github.com/ConservationMetrics/gc-scripts-hub/pull/82, I overlooked that the alerts provider can return multiple rows per month and year when `confidence` values vary (not just when `description_alerts` differs, which is what that PR sought to resolve).

For example:

```
territory_id | type_alert | month | year | total_alerts | description_alerts | confidence
100 | 1 | 10 | 2024 | 5 | bunnies | 1
100 | 1 | 10 | 2024 | 1 | bunnies | 0
100 | 1 | 11 | 2024 | 3 | bunnies | 1
100 | 1 | 11 | 2024 | 2 | bunnies | 0
```

This PR fixes that by including `confidence` towards the hashing of UUIDs, which will guarantee that each of these rows will be written as a unique record. 

I've also adapted the tests to check the `alerts_metadat` table for unique rows with the same month+year but different values for `description_alerts` and `confidence`.

**NOTE:** Much like for #82 I will need to re-generate `alerts_metadata` tables for each user after merging, as this will result in different / new UUIDs. (Hitherto, the scripts would update rows with the same month & year but different confidence values in a random order, which is how I discovered this oversight.)

### DIFF
--- a/f/connectors/alerts/alerts_gcs.py
+++ b/f/connectors/alerts/alerts_gcs.py
@@ -344,9 +344,9 @@ def prepare_alerts_metadata(alerts_metadata, territory_id):
     # Hash each row into a unique UUID; this will be used as the primary key for the metadata table
     # The hash is based on the most important columns for the metadata table, so that changes in other columns do not affect the hash
     filtered_df["metadata_uuid"] = pd.util.hash_pandas_object(
-        filtered_df[["territory_id", "month", "year", "description_alerts"]].sort_index(
-            axis=1
-        ),
+        filtered_df[
+            ["territory_id", "month", "year", "description_alerts", "confidence"]
+        ].sort_index(axis=1),
         index=False,
     )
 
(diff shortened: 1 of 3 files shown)


IamJeffG:
I forget (searched and did not find) where we documented how we are meant to think of "confidence".

https://github.com/ConservationMetrics/gc-scripts-hub/pull/43#discussion_r1899726296 mentions that it "feels to me" like if the data provider retroactively tweaked "confidence" on a row, that's not a new row, that's an update to the existing.
A similar example: Could/would the data provider provide two rows with the same month, year, territory, and confidence?

Or maybe they are using "confidence" as an identifier: like perhaps "we run 3 algorithms side-by-side for our own internal comparisons but you should trust the one we mark as confidence=1, and there will always be exactly one of those" (in which case my confusion lies mostly in the name but whatever)

IamJeffG:
Does anything change retroactively?  If not do we need to add a UUID?

Anyway, I'm fine with this. Please document the heck out of our understanding of these dynamics in the code.

PR #87, opened 2025-04-15
Add Postgres-to-CSV export script and refactor shared export utilities

rudokemper:
## Goal

Add a Postgres-to-CSV export script to our script library, enabling Windmill operators to generate CSV files directly—without needing to query the Postgres database manually.

## What I changed

* Refactored `f/common_logic/db_connection.py` ➜ `f/common_logic/db_operations.py` to include a new function for fetching data from Postgres, in addition to connection logic.
* Created `f/common_logic/save-_disk.py`, which provides a `save_export_file` function that supports different file types (currently: csv, json, and geojson).
* `postgres_to_geojson.py` was updated to use the shared utility functions, and a new script `postgres_to_csv.py` was added using the same pattern.
* Added a test for `postgres_to_csv` using the same CoMapeo mock data as the GeoJSON script.

:eyes:  @nicopace


f/common_logic/db_operations.py
```diff
@@ -0,0 +1,52 @@
...
+
+def conninfo(db: postgresql):
+    """Convert a `postgresql` Windmill Resources to psycopg-style connection string"""
+    # password is optional
+    password_part = f" password={db['password']}" if "password" in db else ""
+    conn = "dbname={dbname} user={user} host={host} port={port}".format(**db)
+    return conn + password_part
+
+
+def fetch_data_from_postgres(db_connection_string, table_name: str):
```
nicopace, line 20: For concistency, I would add db_connection_string: str as in table_name.

f/common_logic/save_disk.py
```diff
@@ -0,0 +1,41 @@
...
+    db_table_name : str
+        The name of the database table, used to name the output file.
+    storage_path : str
+        The directory path where the file will be saved.
+    file_type : str
+        The format to save the file as: "json" or "csv".
+    """
+    storage_path = Path(storage_path)
+    storage_path.mkdir(parents=True, exist_ok=True)
+    file_path = storage_path / f"{db_table_name}.{file_type}"
```
nicopace, line 29: this line can be vulnerable to path traversal attacks if the input strings (db_table_name and file_type) come from an untrusted source and are not properly sanitized.

like:

```
file_path = storage_path / "../../etc/passwd.txt"
```

You mitigate this by:
1. Ensure storage_path is absolute and resolved: `storage_path = storage_path.resolve()` (double check, ChatGPT code)
2. Sanitize inputs
```python
import re

def sanitize_filename(name):
    return re.sub(r"[^a-zA-Z0-9_-]", "_", name)

safe_name = sanitize_filename(db_table_name)
safe_type = sanitize_filename(file_type)
file_path = storage_path / f"{safe_name}.{safe_type}"
```
3. Check final path is within allowed directory

```python
final_path = (storage_path / f"{safe_name}.{safe_type}").resolve()

# Ensure the path is within the storage directory
if not str(final_path).startswith(str(storage_path)):
    raise ValueError("Attempted path traversal detected!")

```

nicopace: A simpler version of this would be:

```python
from pathlib import Path

def get_safe_file_path(storage_path, db_table_name, file_type):
    # Build path safely
    storage_path = Path(storage_path).resolve()
    file_path = (storage_path / f"{db_table_name}.{file_type}").resolve()

    # Check the resolved path stays within the storage directory
    if not file_path.is_relative_to(storage_path):
        raise ValueError("Invalid path: possible path traversal detected.")

    return file_path
```

nicopace (commented):
It looks great Rudo.
Have not run it, did an overview and left some comments.

f/common_logic/save_disk.py
IamJeffG: He does already validate `file_type`, just below.

And also I do like `get_safe_file_path` :arrow_up:

f/common_logic/save_disk.py
```diff
@@ -0,0 +1,41 @@
...
+    Parameters
+    ----------
+    data : list or dict
+        The data to be saved. For CSV, should be a list of rows including a header.
+    db_table_name : str
+        The name of the database table, used to name the output file.
+    storage_path : str
+        The directory path where the file will be saved.
+    file_type : str
+        The format to save the file as: "json" or "csv".
```
IamJeffG, line 25: or "geojson" apparently

f/export/postgres_to_file/tests/postgres_to_csv_test.py
```diff
@@ -0,0 +1,31 @@
...
+
+    main(
+        pg_database,
+        "comapeo_data",
+        asset_storage,
+    )
+
+    with open(asset_storage / "comapeo_data.csv") as f:
+        data = f.read()
+        assert data == (
```
IamJeffG, line 15: This is fine, but also FYI consider a "HEREDOC" to store the expected value.

https://stackoverflow.com/questions/9589301/python-here-document-without-newlines-at-top-and-bottom

IamJeffG (approved):

f/common_logic/save_disk.py
rudokemper: Thanks. I think at some point we could benefit from a repo-wide security audit, informed by any guardrails that Windmill has in place to prevent this kind of thing from happening already. But this being a shared code module, I opted to add this as a best practice :+1:

PR #88, opened 2025-04-16
Add shell script to push changes to multiple workspaces at once

rudokemper:
## Goal

To provide a simple shell script (and documentation) to push changes to multiple workspaces, defined in a `.env` file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,17 +20,31 @@ Some of the tools available in the Guardian Connector Scripts Hub are:
 ![Available scripts, flows, and apps in gc-scripts-hub](gc-scripts-hub.jpg)
 _A Windmill Workspace populated with some of the tools in this repository._
 
-## Deploying the code to a Windmill workspace
+## Deploying the code to Windmill workspaces
 
 [Install the Windmill CLI](https://www.windmill.dev/docs/advanced/cli), and
 [set it up to talk to your deployed Workspace](https://www.windmill.dev/docs/advanced/cli/workspace-management).
 
-Then push the code in this Git repo your workspace:
+Then push the code in this Git repo to your workspace:
 
     wmill sync push --skip-variables
 
 Folders named `./tests/` are excluded by `wmill.yml` from syncing to Windmill —
 because otherwise Windmill tries to make the tests (as with ALL python files) into bona-fide Windmill scripts.
+
+This repo also provides a shell script to batch push changes to a number of workspaces at once. To use this:
+
+1. Ensure the script is executable:
+
+       chmod +x bin/push.sh
+
+2. Populate a list of workspace names in a `.env` file, like so:
+
+       WORKSPACES=gc-windmill,gc-testing-server,...
+
+3. Run:
+
+       bin/push.sh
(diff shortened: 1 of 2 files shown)


IamJeffG:
We do have https://github.com/ConservationMetrics/gc-programs/pull/21 -- tho it hard-codes names.

In my mind, gc-scripts-hub is for one stack at a time, whereas other repos are for the few of us who are deploying many stacks.  What do you think?

rudokemper:
Ah! 

I think the fact that I totally forgot that you had already contributed a script reflects my somewhat different mental model, where I prefer to have as much of the operations associated with the usage of a tool in its main repository as possible. 

Whereas the contents of the `gc-programs` repository is largely stuff that doesn't really fit anywhere, or programmatic files relating to a specific partner. To wit, that `scripts/` dir is largely one-off scripts, so I wouldn't have thought to look there for a Windmill-related script.

But these conventions are not super clear-cut by any means. I think now I will  remember where you placed this script.

README.md
```diff
@@ -20,18 +20,32 @@ Some of the tools available in the Guardian Connector Scripts Hub are:
...
     wmill sync push --skip-variables
 
 Folders named `./tests/` are excluded by `wmill.yml` from syncing to Windmill —
 because otherwise Windmill tries to make the tests (as with ALL python files) into bona-fide Windmill scripts.
 
+This repo also provides a shell script to batch push changes to a number of workspaces at once. To use this:
+
+1. Ensure the script is executable:
+
+       chmod +x bin/push.sh
```
IamJeffG, line 39: You can set this mode in git, so it _will defintely_  be executable going forward.

README.md
```diff
@@ -20,18 +20,32 @@ Some of the tools available in the Guardian Connector Scripts Hub are:
...
+
+2. Populate a list of workspace names in a `.env` file, like so:
+
+       WORKSPACES=gc-windmill,gc-testing-server,...
+
+3. Run:
+
+       bin/push.sh
```
IamJeffG, line 47: In the spirit of "explicit is better than implicit":

We might consider making the user explicitly specify the env vars (and that might be from a `.env` file). So put another way, the script itself does not `source .env`, it only reads environment variables that are set.

Ways of execution would include

    $ WORKSPACES=gcwindmill bin/push.sh

or from any arbitrary env file,  use a subshell (the parens) to avoid keeping these env variables set in your shell beyond this one command.

    $ (source .env ; bin/push.sh)

I think this change would require just removing line 5 of your script, and some documentation changes.

bin/push.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+
+set -e
+
+source .env
+
+IFS=',' read -ra WORKSPACE_LIST <<< "$WORKSPACES"
```
IamJeffG, line 7: Maybe error if none defined.

IamJeffG (commented):
That said: I am 100% fine putting this here.  I am attaching some suggestions if you want to follow this thru.

PR #90, opened 2025-04-20
[comapeo] Pre-scan attachments dir to avoid redundant downloads

rudokemper:
## Goal

Previously, the CoMapeo observations script checked for file existence **before** knowing the MIME type (and thus the final file extension), which caused every attachment to be re-downloaded even if it already existed locally. This fix resolves that by checking against existing stems before making any requests.

This change assumes that the CoMapeo API will never return two of the same stem but with different extensions (e.g. `my-observation.m4a` and `my-observation.jpg`).

## What I changed

* Added `build_existing_file_map()` to scan the CoMapeo project's `attachments/` directory and track all existing files by stem.
* Modified `download_attachment()` to skip download if a file with the same stem already exists (regardless of extension).
* Updated `download_and_transform_comapeo_data()` to use the new file-stem map when checking for already-downloaded attachments

### DIFF
--- a/f/connectors/comapeo/comapeo_observations.py
+++ b/f/connectors/comapeo/comapeo_observations.py
@@ -118,7 +118,7 @@ def fetch_comapeo_projects(server_url, access_token, comapeo_project_blocklist):
     return comapeo_projects
 
 
-def build_existing_file_map(directory):
+def build_existing_file_set(directory):
     """
     Builds a set of existing file names (without extensions) in the specified directory.
     This is used to check for existing files before downloading new ones.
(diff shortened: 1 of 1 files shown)


IamJeffG (approved):
Is there any way to know mimetype from Comapeo before downloading the file? Or the original stem?  In that case we could compute the destination path before downloading, and do less work than this heuristic that compares against all previously downloaded files.  (If not, it might be a nice feature to request from Comapeo)

rudokemper:
> Is there any way to know mimetype from Comapeo before downloading the file? Or the original stem? In that case we could compute the destination path before downloading, and do less work than this heuristic that compares against all previously downloaded files. (If not, it might be a nice feature to request from Comapeo)

Yeah, I agree. I tried to see if a `HEAD` request to retrieve mimetype would work, but comapeo-cloud (running Fastify) isn't set up to support that. I'll go ahead and merge this for now and follow up with Awana about it.

Issue #79, opened 2025-02-14
Create a common DB Writer class to be used by many scripts


rudokemper:
## Feature Request

Right now, `comapeo_observations`, `odk_responses`, `kobo_responses`, `locusmap`, and `alerts_gcs` have a DBWriter class to manage converting a dataset into structured SQL tables and writing them to Postgres using psycopg2.

There is a tremendous amount of duplication of code across these classes; for all but the alerts script, there is little to no difference whatsoever, sans minor differences in logging messages or some specific property like the exact name of the primary key field.

Let's come up with a common DB Writer class that all of these scripts can use as a modular import, stored in `f/common_logic`.

In so doing, we can upgrade to psycopg3 (c.f. https://github.com/ConservationMetrics/frizzle-dagster/issues/29)

PR #93, opened 2025-05-01
Implement a common DB Writer to rule them all

rudokemper:
## Goal

Closes https://github.com/ConservationMetrics/gc-scripts-hub/issues/79. 

This PR introduces a unified `StructuredDBWriter` class for writing structured and semi-structured data into PostgreSQL. 

It consolidates DB write logic across different connectors, while keeping the surface area as small as possible. The biggest change is support for both predefined schema management (e.g., used in `alerts_gcs`) and dynamic schema handling (used in everything else).

## What I changed

* Added a reusable `StructuredDBWriter` class in `f/common_logic/db_operations.py`. The class implements optional support for:
  * mapping tables to preserve original key names
  * automatic schema evolution (when no predefined schema is passed, e.g. for survey data)
  * key sanitization (for nested or illegal field names) and reversal of segments
* Updated all existing connector scripts with a DB write (alerts, KoboToolbox, ODK, GeoJSON) to use `StructuredDBWriter`, and utilizing the above options as needed. 
  * Kobo, ODK, and GeoJSON are almost identical in  swapping in `StructuredDBWriter` for their bespoke writers. The alerts script needed a bit more work to first define the schemas and pass them to the writer.
* Wrote a bunch of new testing assertions to validate that the refactor preserves current behavior.

## What I'm not doing here

Upgrading to `psycopg_v3`. There's already too much else going on in this refactor.

## Other notes

For sanity's sake, I also tested all connector scripts in a (non-user) Windmill instance and confirmed that data is being written in exactly the same way as before.


f/common_logic/db_operations.py
```diff
@@ -50,3 +54,349 @@ def fetch_data_from_postgres(db_connection_string: str, table_name: str):
...
+    StructuredDBWriter writes structured or semi-structured data (e.g., form submissions, GeoJSON features)
+    into a PostgreSQL table. It optionally supports dynamic schema evolution, key sanitization, and column mapping, or can rely on a predefined schema setup.
+
+    Parameters
+    ----------
+    db_connection_string : str
+        PostgreSQL connection string.
+    table_name : str
+        Destination table name for the data.
+    use_mapping_table : bool
```
IamJeffG, line 70: It seems that `use_mapping_table` applies ONLY if `sanitize_keys` is also True. So it's quite misleading.

Possible remedies:
- _always_ sanitize keys, and then _only_ expose the `use_mapping_table` kwarg.  What happens today if sanitize_keys=False and you pass something weird for column name -- and whatever that is, is that what we _want_ to happen?
- keep `sanitize_keys` optional, but if it's set then _always_ use mapping table.  Like, why wouldn't you want that?

(I did not sleuth current usage examples to see if they provide use cases for one or the other; maybe they answer my speculation)


As it is now, here's a failing test case that the class purports to allow but obviously does not:

```python
@pytest.fixture
def mock_db_connection():
    """A dsn that may be used to connect to a live (local for test) postgresql server"""
    db = testing.postgresql.Postgresql(port=7654)
    dsn = db.dsn()
    dsn["dbname"] = dsn.pop("database")
    yield dsn
    db.stop()


def test_mapping_table_creation(mock_db_connection):
    writer = StructuredDBWriter(
        mock_db_connection, "test_forms", sanitize_keys=False, use_mapping_table=True
    )

    submissions = [
        {"_id": "1", "complex.field": "value1"},
        {"_id": "2", "complex.field": "value2"},
    ]
    writer.handle_output(submissions)

    # Check the mapping table contains expected mappings
    mapping_table = f"{writer.table_name}__columns"
    mappings = writer._get_existing_mappings(mapping_table)
    assert "complex.field" in mappings
    assert mappings["complex.field"] == "complex.field"
```

IamJeffG: Given that this PR is a refactor, lets just make a new GH issue to address this later.

IamJeffG (approved):
I notice there is no test of `f/common_logic/db_operations.py` -- other than indirectly by the various connectors.  I will post a new PR to upgrades to psycopg3 and will add a test suite for this in there too.

f/common_logic/db_operations.py
IamJeffG: https://github.com/ConservationMetrics/gc-scripts-hub/issues/98

IamJeffG:
> I notice there is no test of `f/common_logic/db_operations.py` -- other than indirectly by the various connectors. I will post a new PR to upgrades to psycopg3 and will add a test suite for this in there too.

I pushed a commit https://github.com/ConservationMetrics/gc-scripts-hub/pull/93/commits/f71d985c181eb7682c3d0755b804d1736bcdf412 to this branch that adds the test suite I had mind.

rudokemper:
Thanks for the contribution to this!

PR #99, opened 2025-05-20
Upgrade to Psycopg3

IamJeffG:


rudokemper:
https://github.com/ConservationMetrics/gc-scripts-hub/pull/93 was squashed and merged, so this branch will need to be interactively rebased onto main to clean commit history, as I did for the Kobo metadata and Timelapse branches.

Issue #73, opened 2025-02-04
Epic: expanding GC Uploader to support multiple file formats


rudokemper:
https://github.com/ConservationMetrics/gc-scripts-hub/pull/72 introduced a new "GC Uploader" Windmill app to handle Locus Maps exports exclusively.

However, we know that this is not the only type of file that our users will want to upload. Some other ones that we've already heard about include:

* Mapeo or CoMapeo exports (GeoJSON, or ZIP containing GeoJSON and attachments).
* KoboToolbox CSV/XLS export that was manually cleaned up by a user in Excel.
* Compressed files containing Esri Shapefile data.
* Compressed files with Timelapse template and annotation data, including media files.
* CSV or GeoJSON data from other sources.

Rather than creating a separate "GC Uploader" app for each file type, we could adapt the existing application to recognize different formats and schemas, triggering the appropriate connector script automatically.

Perhaps more ambitiously, in the same scope of work we might consider consolidating two or more connector scripts that are broadly similar in flow, and only differ in relation to data structures.

The user experience could be roughly similar to that of [Felt's "Upload Anything"](https://felt.com/blog/upload-anything), providing a single entry point to handle many and diverse file formats that our users want to store in GuardianConnector.

This issue will focus on technical scoping to determine the best approach in Windmill for extending the GC Uploader to support multiple file formats. The issue will be closed once we have converged on a clear approach, at which point implementation work can be scoped and planned separately.

### User stories and inspiration

<details>
<summary>A previously compiled user story</summary>

**Justification**:

Many Indigenous communities already have some data of their own that was created in previous projects, or may work with tools that are not directly part of the GuardianConnector toolkit. For example, they may have GPX data from working with a Garmin GPS, spreadsheet data from working with Excel, or Shapefiles from working with Esri or QGIS.

In addition, most Indigenous communities do not have access to someone, whether a member of the community or an external ally, that knows how to manipulate SQL directly. (In Rudo’s experience, this is a rare skill even among professional NGOs that might have GIS staff.)

The data warehouse is intended to be a centralized hub for all of the community’s data. Hence, communities should be able to upload their external data without needing to work with SQL, or requiring the help of CMI (or other) technical team.

**User story:**

As an Indigenous user of the Biocultural Monitoring System that does not know how to use SQL,

I want to upload my external CSV, XLS, SHP, GeoJSON, KML, or GPX data to the data warehouse,

[...]

PR #109, opened 2025-06-13
Add utility function to detect structured data file types

rudokemper:
## Goal

Towards #73. Adds a function to detect and validate structured tabular or geospatial data types.

As with #108, the idea is to build shared logic that can be used by a Windmill app responsible for uploading files to the data warehouse. Before processing anything, we need to identify the file type and perform basic validation of the content. For now, the function returns "unsupported" for any unrecognized or malformed file; that may change once we build the actual app.

The only slightly annoying bit is detecting CSVs. To avoid blindly trusting .csv extensions or using `csv.reader`, we read up to 3 rows, require at least a header and one row of data, and check that all rows have the same number of columns. A malformed CSV could still bypass this but we can throw an error further downstream once we actually transform or ingest the data, the point here is just detection, not complete validation.


tox.ini
```diff
@@ -41,6 +41,9 @@ commands =
 [testenv:common_logic]
 deps =
     -r{toxinidir}/f/common_logic/db_transformations.script.lock
+    -r{toxinidir}/f/common_logic/db_operations.script.lock
+    -r{toxinidir}/f/common_logic/save_disk.script.lock
+    -r{toxinidir}/f/common_logic/data_validation.script.lock
```
IamJeffG, line 46: if these scripts end up having competing requirements, we could have a problem. For now I'm ok though.

f/common_logic/tests/file_operations_test.py
```diff
@@ -1,6 +1,8 @@
 from pathlib import Path
 
-from f.common_logic.save_disk import save_data_to_file
+from f.common_logic.save_disk import (
+    save_data_to_file,
```
IamJeffG, line 4: is ruff ok with this? https://github.com/ConservationMetrics/gc-scripts-hub/blob/main/CONTRIBUTING.md#ruff-for-python

f/common_logic/tests/data_validation_test.py
```diff
@@ -0,0 +1,86 @@
...
+def test_structured_data_type__valid_csv(tmp_path: Path):
+    file_path = tmp_path / "test.csv"
+    file_path.write_text("col1,col2\nval1,val2")
+    assert detect_structured_data_type(file_path) == "csv"
+
+
+def test_structured_data_type__invalid_csv(tmp_path: Path):
+    file_path = tmp_path / "test.csv"
+    file_path.write_text("key:value,value,more\n1:2:3\nnot,really,a,csv")
+    assert not detect_structured_data_type(file_path) == "csv"
```
IamJeffG, line 33:
```suggestion
    assert detect_structured_data_type(file_path) != "csv"
```

f/common_logic/tests/data_validation_test.py
```diff
@@ -0,0 +1,86 @@
...
+        }
+        """
+    )
+    assert detect_structured_data_type(file_path) == "json"
+
+
+def test_structured_data_type__malformed_json(tmp_path: Path):
+    file_path = tmp_path / "bad.json"
+    file_path.write_text("{ this is not valid json }")
+    assert detect_structured_data_type(file_path) == "unsupported"
```
IamJeffG, line 51: nitpick: I feel this function ought to raise an exception (let's call it `UnrecognizedDataType`). Else it looks like "unsupported" is the name of a valid data type.

It's ok if you reject this idea.

f/common_logic/data_validation.py
```diff
@@ -0,0 +1,152 @@
+import csv
+import json
+import logging
+from pathlib import Path
+
+import pandas as pd
```
IamJeffG, line 6: I notice `openpyxl` in the test requirements (and not in the requirements to run these functions).  Why's it needed there?

Or to take the opposite approach, could we use just `openpyxl` instead of all of pandas to implement these functions?

Reasons to prefer pandas:
- very widely known
- reads easily
- it might (?? I'm guessing) provide a good fallback if openpyxl is not able to be installed.
- it is more in-line with a suggestion of mine down below to consider Fiona for spatial formats

Reasons to prefer openpyxl
- I am guessing pandas installs it already anyway
- more lightweight
- more directly solves your intent
- less likely to end up with version conflicts with other application code, i.e. our scripts are likely to install pandas themselves

f/common_logic/data_validation.py
```diff
@@ -0,0 +1,152 @@
...
+    def is_kml(path):
+        # https://developers.google.com/kml/documentation/kmlreference
+        # We don't care about the schema version, just that it's a KML file.
+        tree = etree.parse(str(path))
+        ns = get_namespace(tree.getroot().tag)
+        return ns.startswith("http://www.opengis.net/kml")
+
+    def is_csv(path):
+        # CSV detection is a bit more complex than the other formats,
+        # because it's overly permissive.
```
IamJeffG, line 101: Why are we ok using `pd.read_excel` but not `pd.read_csv`?

You mentioned in the PR description not wanting `csv.reader` but I am asking you to justify that position.

f/common_logic/data_validation.py
```diff
@@ -0,0 +1,152 @@
...
+    """
+    Detects the type of a structured data file, focusing on geospatial and tabular formats.
+
+    Only files with recognized extensions are considered. The function attempts to
+    parse the file to confirm it matches the expected structure. If the file type is
+    unsupported or the content does not validate, 'unsupported' is returned.
+
+    This function is intended to identify structured datasets used in environmental,
+    geographic, and tabular data pipelines. It supports:
+    - Geospatial formats: geojson, kml, gpx
```
IamJeffG, line 23: in the same way you use pandas to read tabular files, would we consider using [fiona](https://pypi.org/project/fiona/) as a general-purpose reader for vector spatial files? (It will also handle ESRI for you)

I'm going to be suuuuper nitpicky and say that these are all "spatial" files, but not necessarily "geospatial".  (The latter implies a coordinate reference system is specified.)

f/common_logic/data_validation.py
```diff
@@ -0,0 +1,152 @@
...
+    - Tabular formats: csv, xls, xlsx
+    - General structured formats: json
+
+    TODO: Add support for ESRI shapefiles
```
IamJeffG, line 27: heads up this is a multi-file format

IamJeffG (approved):
A bunch of open-ended questions from me :-)

f/common_logic/data_validation.py
```diff
@@ -0,0 +1,152 @@
...
+        return "unsupported"
+
+    expected_type = suffix_map[suffix]
+
+    def get_namespace(tag: str) -> str:
+        if tag.startswith("{"):
+            return tag[1:].split("}")[0]
+        return ""
+
+    # Try validators in priority order; first match wins
```
IamJeffG, line 70: I think this comment isn't right. Your fn doesn't make multiple tries (in any order). It tries the one validator associated with the file extension and either it passes or we return "unrecognized".

So for example if you passed a geojson, with `.json` file extension, it would return "json". That's probably fine (it matches "Only files with recognized extensions are considered" from the docstring)

I'm just saying the code reads in a way that doesnt match this comment.  I'd also recommend considering to group `suffix_map` and `validators` into the same variable (for example a dict from suffix to `(name, val_fn)` tuple).  I think it might read more cleanly.

f/common_logic/data_validation.py
IamJeffG: I think you should ignore my suggestion about Fiona for the time being.  Maybe add a code comment at the top of the file explaining your philosophy. But it's absolutely not worth code changes at this early point.

f/common_logic/data_validation.py
rudokemper: Good catch. Comment was leftover from a working idea that was abandoned.

f/common_logic/tests/data_validation_test.py
rudokemper: Similar to my comment about [error handling](https://github.com/ConservationMetrics/gc-scripts-hub/pull/110#discussion_r2174034313) in #110, I need to take the time to recall how this can work across a Windmill app.

If this wasn't written with usage in a Windmill app in mind, I would totally agree with this. And it does raise a question around how somebody might use this "common logic" function in a different context.

I'm going to make an explicit note to revisit this as part of the work proposed in https://github.com/ConservationMetrics/gc-scripts-hub/issues/73.

f/common_logic/data_validation.py
rudokemper: So, the reason we need `openpyxl` in the test requirements is because we use pandas `df.to_excel` which uses `openpyxl` as an engine for writing, and as it turns out, that does need to be installed separately. So without having this in `requirements.txt`, the one test which leverages that method will fail with a ModuleNotFoundError.

https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.to_excel.html

(Turns out you can also use `xlsxwriter` as an engine, but that requires installation too.)

I will add an inline comment to clarify this in requirements.

I'll also make a note to clarify the current approach, and possibilities of revisiting (and also using `fiona`) in the future.

rudokemper: You are right. Fixing this.

Issue #97, opened 2025-05-19
Add script to extract and transform uploaded Auditor2 data


rudokemper:
## Feature Request

Add a script that takes a path for Auditor2 data (in ZIP file format), transforms the data to write to a database, and copies attachments to the datalake.

See work towards defining a data schema for Auditor2 exports [here](https://github.com/ConservationMetrics/gc-scripts-hub/issues/68), and a branch with a fixture for testing [here](https://github.com/ConservationMetrics/gc-scripts-hub/tree/auditor2/f/connectors/auditor2/tests/assets).

See also [work and discussion towards creating a script for ingesting camera trap data (raw and annotations) from Timelapse](https://github.com/ConservationMetrics/gc-scripts-hub/issues/71) which will largely parallel, and inform, this script.

abfleishman:
@rudokemper yes I think that timeline is reasonable.  I will look at these csvs today.  Specific changes I am planning: 
reef_type ->habitat
add col `datetime`
review for assuring column names are "generalized" (not reef project specific)

abfleishman:
OK I looked at the csvs and changed the following: 

- added `datetime` column to the `moorea_labels_20250605.csv`
- changed the `Reef_Type` column to `habitat` in `moorea_sites_20250605.csv`
- changed all the column names to `snake_case` (CMI style guide best practices).

I made thees changes in both my original wrapup script and in the csv files on the auditor2 branch see 0de41cc  
The files are here: 
~\CMI Dropbox\CMI_Team\Analysis\2023\NPT_UCSB_Moorea_2023\Clips\NPT_UCSB_Moorea_2023_R1_FLAC

"moorea_deployments_20250605.csv"
"moorea_labels_20250605.csv"
"moorea_sites_20250605.csv"
"moorea_sound_file_summary_20250605.csv"

Issue #106, opened 2025-06-10
[timelapse / a2] Support large file uploads for ETL script access


rudokemper:
## Feature Request

As previously noted [here](https://github.com/ConservationMetrics/gc-scripts-hub/issues/73#issuecomment-2641304421) and [here](https://github.com/ConservationMetrics/gc-scripts-hub/pull/94#discussion_r2135962776), we need to establish a process for uploading very large archive files to cloud storage and enabling access to those files by the Timelapse and Auditor2 ETL scripts (which does not exist yet - https://github.com/ConservationMetrics/gc-scripts-hub/issues/97). 

Using the [file input](https://www.windmill.dev/docs/apps/app_configuration_settings/file_input) in a Windmill app will not do for large files, so we'll likely need to handle uploads outside of Windmill. For example, we could upload files to Azure Blob Storage and then use the Python Azure Storage SDK within the scripts to access the files.

At this point, it's unclear whether support for Azure alone will be sufficient, or whether we should aim to support other cloud storage providers (e.g., S3, GCS) as well. The workflow should ideally be flexible enough to accommodate different backends if needed. But given that CMI is likely to be the only user of these scripts in the short-to-medium term, it's fine to start with Azure only. 

In terms of the actual workflow: do we just upload files to blob storage manually, and provide whatever path and credentials are needed for the Python SDK when running or scheduling a script in Windmill? @IamJeffG I'm interested to learn what you've done in the past (feel free to just link to a code snippet if that's easiest).

To close this issue, let's:

- [x] Agree upon the workflow 
- [x] Modify the scripts accordingly
- [ ] Provide necessary documentation

rudokemper:
Some kind of process to delete a file (either usage of SDK with a write permission credential, or just a reminder to delete the file)

rudokemper:
[Workspace object storage](https://www.windmill.dev/docs/core_concepts/object_storage_in_windmill#workspace-object-storage) does support Azure Blob containers, but there are in fact limitations on the free plan that will prevent this from being useful to us:

> Windmill S3 bucket browser will not work for buckets containing more than 20 files and uploads are limited to files < 50MB. Consider upgrading to Windmill [Enterprise Edition](https://www.windmill.dev/pricing) to use this feature with large buckets.

I have verified that this is the case:

<img width="1535" height="684" alt="Image" src="https://github.com/user-attachments/assets/5afc08cf-a39d-44e8-a10d-7dd053139cd5" />

[...]

Issue #102, opened 2025-05-28
[timelapse] Add script params and apply transformations as needed by front end


rudokemper:
## Feature Request

https://github.com/ConservationMetrics/gc-scripts-hub/pull/94 adds an open-ended Timelapse ETL script, and https://github.com/ConservationMetrics/gc-explorer/issues/46 describes a user interface for visualization.

In order to meet the front-end needs, we should figure out what kind of transformations we need to apply to Timelapse data during ETL.

We're not ready to do this until we better understand what exactly the front end will need. But filing the issue for posterity.

rudokemper:
Is it sane to even pursue this, given the highly dynamic nature of Timelapse database tables, constructed from user-defined folder hierarchies and field schemas? Take lat/long for example...

The frontend will need coordinates in order to render map points. However, because Timelapse projects can define an arbitrary number of folder levels (i.e. `Project → Region → Station → Deployment`... or `Project → Village → Station → Camera Trap`... or `...`) and rename any field, it's not possible to know which level contains geospatial coordinates, or what the coordinate columns are named. This location data could live in any level, under any name, depending on how the Timelapse project was configured by the user. There is no guaranteed schema or naming convention.

The approach suggested in this issue is to add script parameters to avoid any failure-prone heuristic attempt to detect lat/long columns. Fine. But even for lat/long, the script user will need to provide (1) the folder level and (2) the names of the lat/long columns. Let's say the folder level is "Station". To determine which _actual_ database table that is an alias for — recall that the folder database tables are simply named `Level{n}`, and this is not something the users ever sees in the Timelapse software — we have to first refer to the `FolderDataInfo` lookup table, which maps level numbers to aliases. So, we could have an input param `folder_level_alias_containing_lat_long`, where the user provides the alias (e.g. "Station"), and we resolve it to a level number via `FolderDataInfo`. Then `lat_column_name` and `long_column_name` params would define the fields to extract. 😮‍💨

I wonder: might it be more flexible to shift this responsibility to the frontend, where users can configure which level and columns should be used for mapping? There, we can leverage the `FolderDataInfo` lookup table to show human-friendly alias names directly, and potentially create a wizard-style flow, e.g. a modal asking "Which of your folder levels contains coordinates?" followed by "Select your latitude and longitude columns."

[...]

PR #111, opened 2025-06-19
Add script to extract and transform uploaded Auditor 2 data (first pass)

rudokemper:
## Goal

Closes https://github.com/ConservationMetrics/gc-scripts-hub/issues/97.

## What I changed

* Added ETL script and tests for Auditor 2 data, based on @abfleishman's fixtures.
* This script follows many of the same patterns as other `connector` scripts (in particular the Timelapse one, which also works with a ZIP file), but I want to highlight the following aspects:
  * To avoid collisions with previous Auditor 2 imports in the data warehouse, I added a function (leveraging the new [check if table exists](https://github.com/ConservationMetrics/gc-scripts-hub/pull/108) module) to check if the `project_name` has already been used to write Auditor 2 tables in the database; if so, it raises an error.
  * We explicitly check for the five expected CSVs and raise an error if any are missing.
  * For data transformation, the only operations performed are constructing an `_id` field (for the SQL primary key) and adding `g__coordinates` and `g__types` columns to the `sites` data because [unlike Timelapse](https://github.com/ConservationMetrics/gc-scripts-hub/issues/102), we *can* deterministically identify which fields store this information in Auditor 2 exports.

## What I'm not doing here

* Any further transformation work, pending further articulation of front end requirements.
* Handling how the Auditor 2 ZIP file "gets there". For now, the script assumes it already exists in volume-mounted storage. How that will actually happen (for Auditor2 or Timelapse) will be handled in https://github.com/ConservationMetrics/gc-scripts-hub/issues/106


rudokemper:
BTW -- why now? To wrap up some WIP on this from earlier in the year, and so we can cite it as progress towards our active grant supporting this work, in a forthcoming report...

f/connectors/auditor2/README.md
```diff
@@ -0,0 +1,32 @@
+# Auditor 2: Import Annotated Acoustic Data
+
+This script processes annotated acoustic data from an [Auditor 2](https://github.com/ConservationMetrics/AuditorServer) project, via a ZIP file containing 5 expected CSVs and attachments (audio and spectrogram images).
+
+The script expects the following five CSV files, each identified by specific substrings in their filenames:
+- `deployments`: Must include a `deployment_id` field.
+- `human_readable_labels`: No specific ID field required.
+- `labels`: No specific ID field required.
+- `sites`: Must include a `site_id` field.
+- `sound_file_summary`: Must include a `deployment_id` field.
```
IamJeffG, line 10: add .csv extensions to the filenames.

f/connectors/auditor2/auditor2.py
```diff
@@ -0,0 +1,234 @@
...
+    project_name: str,
+    delete_auditor2_zip: bool = True,
+    attachment_root: str = "/persistent-storage/datalake",
+):
+    raise_if_project_name_exists(db, project_name)
+
+    storage_path = Path(attachment_root) / "Auditor2" / project_name
+    auditor2_zip_path = Path(auditor2_zip)
+
+    actual_storage_path = extract_auditor2_archive(auditor2_zip_path, storage_path)
```
IamJeffG, line 32: extract_auditor2_archive returns None, so this will not work right

f/connectors/auditor2/auditor2.py
```diff
@@ -0,0 +1,234 @@
...
+
+    for file in csv_files:
+        for key in required_keys:
+            if key in file.stem:
+                with file.open("r", encoding="utf-8") as f:
+                    reader = csv.DictReader(f)
+                    rows = list(reader)
+                auditor2_tables[key] = rows
+                found_keys[key] = True
+                break
```
IamJeffG, line 161: I notice these are substrings, so could produce subtle bugs when you grab one.
For example, say the ZIP contains `foo_sites_20250505.csv` and `bar_sites_20250505.csv`.

Consider adding a check that each required_key is found exactly once.  One idea, you could use a `from collections import Counter` for `found_keys` instead of `dict`, and then later assert all values == 1.  Or inside these `for` loops you can just check if `found_keys` is already set, but in this case please make it a `set()` instead of dict.

f/connectors/auditor2/auditor2.py
```diff
@@ -0,0 +1,234 @@
...
+    transformed_data = {}
+
+    # Process each table in the Auditor 2 data
+    for table_key, rows in auditor2_data.items():
+        table_name = f"auditor2_{project_name}_{table_key}"
+        id_field = id_fields.get(table_key)
+
+        for index, row in enumerate(rows):
+            if id_field:
+                row["_id"] = row.get(id_field, "").strip()
```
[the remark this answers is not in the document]
rudokemper: This was a bug with no real explanation behind it. I attribute it to coding (too) late at night.. sorry, and fixed.

IamJeffG (approved):

Issue #95, opened 2025-05-12
Add a script to download all (or most?) data, aka "exit plan"


rudokemper:
## Feature Request

We have committed to making it possible for users to "download all of their data" on GuardianConnector. The primary reasons given for this include the possibility that they would like to stop using the stack but retrieve their data before doing so, or if they just want to do a backup at any given time. This is an important feature towards guaranteeing data sovereignty.

In what follows, I will first provide some background on what we might mean, and what our users might expect, from downloading their data. I will then focus on the Windmill / `gc-scripts-hub` specific tasks.

> [!NOTE]
> _At this time of writing, this is not a priority issue. I am just documenting some ideas that came to me this morning. Thoughts and reflections on anything here are welcomed._

### What data are we talking about?

What does “download all of their data” actually mean? Here’s a breakdown of the relevant data types:

1. **Tabular data** stored in the `warehouse` Postgres database (e.g. data ingested by connector scripts like Kobo or CoMapeo, or uploaded manually)
2. **Files** in `/persistent-storage/datalake` (e.g. attachments from connector scripts, file exports like GeoJSON, or files added via Filebrowser)
3. **GuardianConnector Explorer** and **Superset** config (`guardianconnector` and `superset-metastore` DB tables)
4. **Windmill** config for scheduled scripts (`windmill` db table)
5. **Windmill** run history, stored in windmill logs on `captain--windmill-worker-logs` volume
6. **CapRover** platform config (`/captain/data/config-captain.json`)
7. **CoMapeo Cloud** raw data (stored in a Docker volume `captain--comapeo-data`)
8. **auth0** authentication and session history (stored externally)
9. _...am I forgetting anything? Probably!_

(Note: Third-party resources like Mapbox or Planet are not considered here.)

### User stories for downloading data

But what might our users actually want? I think there are potentially three user stories:

#### **User story A**: I’m done with GuardianConnector and want to take my project data with me.

This is likely the most obvious user story. For this, data types (1) and (2) are the most essential. But perhaps we could also offer the option to download (4) and (5) so that the GC connector config -- both upstream (ETL scripts) and downstream (front ends) -- is provided as a source of reference for how the data was ingested and visualized, in case it's helpful for other work.

#### **User story B**: I want to pause my use of GuardianConnector, but might return later.

[...]

PR #112, opened 2025-06-24
Add script to download all Postgres warehouse data, and "exit plan" documentation

rudokemper:
## Goal

Towards https://github.com/ConservationMetrics/gc-scripts-hub/issues/95.

This PR contributes an initial "download all of my project data / exit plan" implementation to support data sovereignty and project offboarding needs. Specifically, it adds a script to export all tables from the `warehouse` Postgres database as CSVs, bundled into a downloadable ZIP archive.

## What I changed

* Added `download_all_data_from_postgres.py` script to export all tables from the warehouse DB. Each table is exported as a CSV and bundled into a single ZIP.
* Added a `fetch_tables_from_postgres()` utility function to `db_operations`.
* Added tests for both.
* In the directory `README.md`, I included a good amount of the discussion about what we currently mean, and don't mean, by "exit plan".

## What I'm not doing here

* Not including datalake file attachments. That can be a separate script, and the two can be connected in a Windmill Flow. Even more importantly, I think we need to think through where the output would be stored (e.g. maybe make it a requirement to stream to outside object storage?), since a ZIP file of all content in `/persistent-storage/datalake/`... to be stored in `/persistent-storage/datalake/` itself... would double the disk usage and potentially break the instance!


IamJeffG:
Before we continue I'd request we at least take a look at other tools to do the dump of each table you've enumerated, to solve this in a more robust way, and with fewer lines of code that we need to maintain.  (I suspect we probably will need to keep what you have so far, about enumerating tables)

* pg_dump: I know pg_dump doesn't directly output SQL, it outputs either (text file of SQL commands) or a `.dump` file.  Maybe [there's a way to restore the dump file straight to CSV?](https://stackoverflow.com/questions/18114333/how-to-extract-a-table-from-a-dump-file-into-a-csv)
* psql `\copy` commands: We can drive them with a small Python script and emit CSVs per table. Here `psql` is the command-line client, not psycopg (though I guess psycopg has copy ability too)
* Since we're already running in Python land, Panda's `read_sql_table + to_csv` could be absolutely the most expedient.

rudokemper:
Thanks. You're right that the way I've done this so far is not the most efficient. (And please don't review further until we decide what to do.)

In addition to what you mentioned, it does seem that we could also use psycopg's [`copy_expert`](https://www.psycopg.org/docs/cursor.html#cursor.copy_expert) as an interface to Postgres `COPY`.

I would prefer to use either psycopg or pandas, since we use these two libraries extensively across the codebase already, and we can stay working in a full Python environment. (I might also refactor `postgres_to_csv` itself to use either approach, even if we don't end up using that function here any longer.) 

Insofar as the task has first-class support in psycopg (including in v2), I think I would prefer to use `copy_expert` for simplicity and to leverage native Postgres command functionality.

What do you think?

f/export/postgres_to_file/postgres_to_csv.py
```diff
@@ -1,20 +1,52 @@
...
+
+from f.common_logic.db_operations import conninfo
 
 logging.basicConfig(level=logging.INFO)
 logger = logging.getLogger(__name__)
 
 
 def main(
-    db: postgresql,
+    db: dict,
```
IamJeffG, line 13: I think you could keep this `db : postgresql`

by importing it above:

    from f.common_logic.db_operations import conninfo, postgresql


And that keeps typecheckers most useful.

IamJeffG (approved):

PR #122, opened 2025-07-18
Implement use of Fiona for spatial data conversion

rudokemper:
## Goal

Stacked onto https://github.com/ConservationMetrics/gc-scripts-hub/pull/121. Refactors `gpx_to_geojson()` and `kml_to_geojson()` to use Fiona to avoid complex manual XML parsing, and benefit from its GDAL/OGR drivers.

For GPX files, the implementation now uses `fiona.listlayers()` to read all available layers (waypoints, tracks, routes, track_points) and processes them with Fiona's native GPX driver. 

For KML files, a hybrid approach was still necessary because Fiona's KML driver (which requires manual enabling via `fiona.supported_drivers['KML'] = 'rw'`) apparently only reads basic elements like Name and Description, but ignores ExtendedData elements that contain critical properties. So some of the previous XML parsing is still in place, but I think it reads a lot better overall.

I also added some additional KML/GPX/GeoJSON fixtures from Overpass, and went a little overboard with tests for them, but they pass.

Sorry this is messy between the two draft PRs, but I felt motivated to try Fiona!


f/common_logic/data_conversion.py
```diff
@@ -206,85 +209,47 @@ def json_to_csv(path: Path):
...
+                links = []
+                for field in sorted(link_fields):  # Sort for consistent output
+                    if properties[field]:
+                        links.append(str(properties[field]))
+                        del properties[field]
+
+                if links:
+                    properties["link"] = ", ".join(links)
+
+                # Clean up empty properties
```
IamJeffG, line 243: is this an artifact of reading GPX  (i.e. GPX or fiona adds empty properties that aren't real), or is this business logic?


I think this, and perhaps also the comma-joining links, are a lossy transformation that might be noted in the docstring

IamJeffG (approved):
Did you have to do anything to make sure the OGR drivers (which fiona uses behind the scenes) are installed?

If the world hasn't changed since I last used Fiona in 2015, then we need to address installing OGR for
1. local dev (in README)
2. CI
3. deployment docker image

rudokemper:
> Now that you tried it, what do you think?

I also wish that it would have brought the line count down more, and I ultimately wasn't 100% successful in getting rid of XML parsing altogether. But ultimately, I feel so much more confident deferring to this library (and indeed OGR) for format conversions. That enough makes me happy with it.

> Did you have to do anything to make sure the OGR drivers (which fiona uses behind the scenes) are installed?

So I read online that it's quite common to need to install OGR drivers to use Fiona effectively, but that it's also possible to just do `pip install fiona` which installs some though not all drivers. https://pypi.org/project/fiona/ has some cautionary language that the PyPi package will only suffice for simple projects as not everything is included. That said, all tests seemed to work fine in the tox environment (I have not tried on Windmill runtime). So this might be sufficient?

If not, I honestly wouldn't mind if we installed requisite drivers in the Docker image (though spatial data libs can be heavy). `gc-scripts-hub` deals with enough spatial data that Fiona could be useful in a number of other places.

IamJeffG:
With Windmill there is no Docker image to which we can add OGR for the workers. It's worth testing fiona on a couple files here.

rudokemper:
Tested this in Windmill runtime. I was missing some requirements, but after adding them, Fiona ran like a charm :+1: 

And discovered a bug in so doing, which I fixed and am now testing for.

Issue #129, opened 2025-08-07
Change `data_source` for alerts provider in `alerts_gcs` script.


rudokemper:
## Feature Request

c.f. https://github.com/ConservationMetrics/gc-scripts-hub/blob/8574d48e789d88dbec4bb4a80b6f6791f71457c3/f/connectors/alerts/alerts_gcs.py#L380-L382

I wonder if it's best to just add a script parameter that allows the user to set the data source.

IamJeffG:
Yes I agree to add a script parameter. The windmill schedule can set that value, so in practice the user still rarely has to deal with it.

PR #133, opened 2025-08-12
Allow user to set `alerts_provider`; don't override existing provider

rudokemper:
## Goal

Closes https://github.com/ConservationMetrics/gc-scripts-hub/issues/129.

To address this, I needed to introduce a fix to ensure that existing records are kept intact with the previously hard-coded provider.

Let's assume we had an alert with `data_source`="first_provider", and now we set `alerts_provider`="second provider".

**Before Fix:**
* Same alert + any provider → UUID: abc123
* New records get `data_source`="second_provider"
* Existing records _also_ get updated with `data_source`="second_provider"

**After Fix:**
* Same alert + any provider → UUID: abc123
* New records get `data_source`="second_provider"
* Existing records keep `data_source`="first_provider"

(And to be explicit: the reason we don't factor providers into hashing UUID's is to ensure that no records are duplicated if the alert provider changes.)

## What I changed

* Allow the Windmill user to set a parameter `alerts_provider` instead of hard-coding it as `data_source`
* Added a helper function `_get_existing_data_sources()` to retrieve existing `data_source` values for given alert IDs (hashed as UUID) from the DB.
* In `prepare_alerts_metadata()`, compare incoming `data_source` (set via `alerts_provider` param) with existing database values. If an existing record is found, preserve its original `data_source` value. Only new records get the current `alerts_provider` value.
* Added tests to verify this behavior.
* Updated the Alerts Flow as well.

### DIFF
--- a/f/connectors/alerts/alerts_gcs.py
+++ b/f/connectors/alerts/alerts_gcs.py
@@ -385,7 +385,6 @@ def prepare_alerts_metadata(alerts_metadata, territory_id, alerts_provider):
         index=False,
     )
 
-    # TODO: if this changes for future alerts, we will need to ensure that existing records are not overwritten.
     filtered_df["data_source"] = alerts_provider
 
     # Replace all NaN values with None
--- a/f/connectors/alerts/alerts_gcs.py
+++ b/f/connectors/alerts/alerts_gcs.py
@@ -33,6 +33,7 @@ logger = logging.getLogger(__name__)
 def main(
     gcp_service_acct: gcp_service_account,
     alerts_bucket: str,
+    alerts_provider: str,
     territory_id: int,
     db: postgresql,
     db_table_name: str,
@@ -49,6 +50,7 @@ def main(
     return _main(
         storage_client,
         alerts_bucket,
+        alerts_provider,
         territory_id,
         db,
         db_table_name,
@@ -59,6 +61,7 @@ def main(
 def _main(
     storage_client: gcs.Client,
     alerts_bucket: str,
+    alerts_provider: str,
     territory_id: int,
     db: postgresql,
     db_table_name: str,
@@ -71,6 +74,8 @@ def _main(
     storage_client : google.cloud.storage.Client
     alerts_bucket : str
         The name of the GCS bucket containing alerts.
+    alerts_provider : str
+        The name of the alerts provider.
     territory_id : int
         The ID of the territory for which alerts are being processed.
     db : postgresql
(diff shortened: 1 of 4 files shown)


f/connectors/alerts/alerts_gcs.py
```diff
@@ -96,11 +101,14 @@ def _main(
...
+    alerts_metadata_table_name = f"{db_table_name}__metadata"
     prepared_alerts_metadata, alerts_statistics = prepare_alerts_metadata(
-        alerts_metadata, territory_id
+        alerts_metadata, territory_id, alerts_provider, db, alerts_metadata_table_name
     )
 
-    prepared_alerts_data = prepare_alerts_data(destination_path, geojson_files)
+    prepared_alerts_data = prepare_alerts_data(
+        destination_path, geojson_files, alerts_provider
+    )
```
IamJeffG, line 111: I don't recall the distinction between the `alerts` vs `alerts__metadata` tables... but it appears that `data_source` is a column in both of them. So why does `prepare_alerts_data` not also have to do the same disambiguation of which rows are already in this table vs new (using `_get_existing_data_sources`)?

f/connectors/alerts/alerts_gcs.py
```diff
@@ -370,16 +426,25 @@ def prepare_alerts_metadata(alerts_metadata, territory_id):
...
-    filtered_df["data_source"] = "terras"
+    # Preserve existing data_source values for existing records
+    existing_data_sources = _get_existing_data_sources(
+        db_conn_info, table_name, filtered_df["_id"]
+    )
+
+    # Set data_source: preserve existing values, use new provider only for truly new records
+    def get_data_source(row):
+        record_id = str(row["_id"])
+        return existing_data_sources.get(record_id, alerts_provider)
```
IamJeffG, line 445: This isn't quite the same as "new rows will get `alerts_provider`, old rows remain unchanged". Rather it's "replace all false-y values with `alerts_provider`".

So the difference is in the (unexpected) case that there is an old row with null value.

It's fine in this case but it does have a smell like we are doing something wrong.

See my top-level comment for bigger-picture thoughts.

rudokemper:
I fully agree with all of your concerns, from the opaqueness down to whether StructuredDBWriter is doing the right thing.

That said, I thought about this on my run just now, and I'm glad you came to the same conclusion:

> Why do we need subsequent runs of this script to see the old CSVs from the old provider? We can either branch off old files into a different storage account or -- frankly -- delete them or move to a backup location. So script runs only see CSVs from one provider at a time. This seems simplest across the board.

Yes. We can do this. Let's go this route.

I will revise this PR to simply accept the param, and be done with it. And separately discuss with Mayana how to manage the new bucket.

rudokemper:
Done. @IamJeffG can you re-review? Should be non-controversial.

IamJeffG (approved):

Issue #117, opened 2025-07-14
Build a Windmill app for GC File Uploader


rudokemper:
## Feature Request

Per [this architecture diagram](https://gist.github.com/rudokemper/87795465738ba4c477ace6a572f3a046), after all of the core scripting work is completed, it's time to build the Windmill app that ties it all together.

TODO:

- [ ] Structure a [Windmill app](https://www.windmill.dev/docs/getting_started/apps_quickstart) with the desired screens, flows, and state management shown in the above diagram and example flows.
- [ ] Ensure all shared utility scripts and tool-specific transformation logic works well with the Windmill app.
- [ ] Fold the existing [Locus Map](https://github.com/ConservationMetrics/gc-scripts-hub/tree/main/f/connectors/locusmap) and [GC Uploader](https://github.com/ConservationMetrics/gc-scripts-hub/tree/main/f/apps/gc_uploader.app) prototype app (only built to work with Locus Map) into this effort.

rudokemper:
@IamJeffG I fixed the bugs that came up today, and I'd love to pass the app over to you for test-driving sometime (hence assigning you for visibility on the Kanban). It is deployed on `demo` Windmill. But no rush; this can be for next sprint.

rudokemper:
Thanks @IamJeffG.

I will nevertheless split out several PRs that are changes or additions that are not inherently related to the app itself, and would be good to keep as a separate commit -- like a "CSV to Postgres" script that I will contribute momentarily.

IamJeffG:
@rudokemper am I missing anything you need from me on this?

rudokemper:
@IamJeffG not if you're done reviewing the PR! 

(I did see you approve a few days ago, but briefly wondered if you were _truly done_ considering the size of the project. 

Not that I'm at all doubting that you did a thorough review, but moreso anticipating some oversights on my behalf. That said, I did smith away at it for weeks... so one would hope it'd be in relatively good shape 🙂)

Thanks; I'll merge!

IamJeffG:
In fact, I did not do a thorough review and decided that it was not likely to happen, and that having seen it empirically work is going to be good enough :-)  But  yes, I approve.

PR #136, opened 2025-08-20
Generate unique ID if not present in GeoJSON feature

rudokemper:
## Goal

Towards https://github.com/ConservationMetrics/gc-scripts-hub/issues/117. 

Until now, we assumed all GeoJSON features came with an existing `id`, since most were pre-processed by connectors like CoMapeo or Kobo that assign one. Further downstream, `StructuredDBWriter` relies on this `id` when inserting records into the database.

With the Dataset Importer app, we may receive raw GeoJSON uploads where features lack an `id`. This PR fixes that by generating a stable UUID for each feature, derived from a hash of its content.

### DIFF
--- a/f/connectors/geojson/geojson_to_postgres.py
+++ b/f/connectors/geojson/geojson_to_postgres.py
@@ -57,6 +57,9 @@ def transform_geojson_data(geojson_path):
 
     transformed_geojson_data = []
     for i, feature in enumerate(geojson_data["features"]):
+        # TODO: consider using a more deterministic ID generation method,
+        # once we are ready to tackle the challenge of appending data
+        # to existing tables with existing IDs
         feature_id = feature.get("id", str(uuid.uuid4()))
 
         transformed_feature = {


IamJeffG:
I see the intent, but I’m not convinced we can actually define equality today for future objects in a way that's robust. If one feature has `{"key": null}` and another simply omits that key, should we treat them as equal? What about extra metadata fields? My worry is that this becomes a minefield of subtle mismatches.

If we really want "magic" deduplication across uploads, should it come with explicit user confirmation, or at least configurable heuristics (e.g. per dataset)?  I suspect (without evidence) that hashing might not even be the right mechanism in the first place, that it will require more nuance.

By merging this now, I fear we risk baking-in brittle assumptions at the lowest level and this will lead to surprising bugs later.

rudokemper:
Fair enough.

We're not about to support appends to existing datasets anytime in the near (read: in scope) future, so let's leave this as a future thing to resolve.

If you are agreed, I will implement one of your two proposed approaches who do not bake in any questionable assumptions. And I can leave a TODO comment to help us re-orient towards this problem later.

IamJeffG:
Yes, that's probably for the best. Sorry about that!

More realistic use case (from chat) that would break this:

> What if I'm a user and I made the tiniest of changes to an attribute table. Like, I changed a <NULL> value to "" (easy to do in QGIS by editing a cell, and then erasing the contents when you changed your mind). Now when I upload it, it's a duped feature.

Issue #140, opened 2025-09-04
[alerts / geojson] Support for `GeometryCollection`s


rudokemper:
## Feature Request

@maypontes reported that some alerts are coming in with geometries defined as `GeometryCollection`. These can contain multiple geometry types (e.g. `LineString` and `Polygon`). Example:

```json
{
  "type": "FeatureCollection",
  "features": [
    {
      "type": "Feature",
      "geometry": {
        "type": "GeometryCollection",
        "geometries": [
          {
            "type": "LineString",
            "coordinates": [
              [-100.11945344942263, 25.327808970020767],
              [-100.11954328095104, 25.327808970020767]
            ]
          },
          {
            "type": "Polygon",
            "coordinates": [
              [
                [-100.11945344942263, 25.327808970020767],
                [-100.11945344942263, 25.328078464606003],
                [-100.11954328095104, 25.328078464606003],
              ]
            ]
          },
        ]
      },
      "id": "-123456+78910",
      "properties": {
        ...
      }
    }
  ]
}
```

Currently, our scripts write these as a singular record in the database, with `g__type = GeometryCollection` but `g__coordinates = null`, since `geojson_to_postgres` expects `geometry["coordinates"]` (not present for `GeometryCollection`).

### Considerations

There are a few nuances here...

1. **`GeometryCollection`s are valid GeoJSON per [the GeoJSON spec (RFC 7946)](https://datatracker.ietf.org/doc/html/rfc7946?utm_source=chatgpt.com#section-3.1.8)**. So, we should improve `geojson_to_postgres` to handle them gracefully.

2. **However, many clients, including Mapbox/Maplibre GL JS and QGIS, cannot render `GeometryCollection` directly.** For Mapbox, [it's apparently been requested](https://github.com/mapbox/vector-tile-spec/issues/110) before, but to no avail. @maypontes reports that they do not work in QGIS. The usual solution is to split each geometry into its own feature.

    Given that that Mapbox/Maplibre GL JS and QGIS are _the_ primary tools into which our processed GeoJSON features are loaded, I think that we should split the `GeometryCollection` during ETL, instead of forcing that burden to be confronted downstream by the client or end user, as @maypontes affirms this can be a major pain point.

    That said, I do recognize there is a tension here between two design principles of Guardian Connector: **accessibility** and **data warehousing** (specifically, strict fidelity / lossless copying). Curious to @IamJeffG's take on this.

[...]

PR #144, opened 2025-09-16
Add graceful error handling for CoMapeo alerts posting failures

rudokemper:
## Goal

This PR implements error handling for the CoMapeo alerts connector to gracefully handle API posting failures (such as when the API rejects an alert due to having null coordinates or `GeometryCollection` type). 

In particular, when the CoMapeo API rejects an alert with a 400 error (for example, due to null coordinates or an unsupported GeometryCollection geometry type), the script will no longer fail outright. Instead, failed alerts are collected during execution, and a RuntimeError is raised at the end of the run if any alerts could not be posted, similar to the approach for handling failures in fetching attachments in https://github.com/ConservationMetrics/gc-scripts-hub/pull/11.

The reason for making this change is that some alerts with invalid or unsupported geometry have made their way into partner alerts tables (see [#140](https://github.com/ConservationMetrics/gc-scripts-hub/issues/140)
). While we are fixing that issue upstream, this update ensures that partners can still access valid alerts in CoMapeo as they prepare for fieldwork, rather than losing access entirely due to one or more problematic records.

### DIFF
--- a/f/connectors/comapeo/comapeo_alerts.py
+++ b/f/connectors/comapeo/comapeo_alerts.py
@@ -50,7 +50,12 @@ def main(
 
     transformed_unposted_alerts = transform_alerts(unposted_alerts)
 
-    post_alerts(comapeo_alerts_endpoint, comapeo_headers, transformed_unposted_alerts)
+    alerts_failed = post_alerts(
+        comapeo_alerts_endpoint, comapeo_headers, transformed_unposted_alerts
+    )
+
+    if alerts_failed:
+        raise RuntimeError("Some alerts failed to post.")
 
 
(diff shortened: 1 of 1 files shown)


IamJeffG (approved):
Generally I'd think we may want the outer `main` to have some sort of info about what went wrong: You might consider having `post_alerts` return either (or both):
- a `dict` showing how many alerts were successful vs failed
- a `list` of failures (either the exception objects themselves or some summary of each)


Of course if `main` really isn't going to do anything with that info, then what you have is fine.  I think that's justifiable especially in a Windmill script like this, where most humans are going to have `post_alerts`'s warning logs just as equally accessible to them as `main`'s return value.

Issue #145, opened 2025-09-16
[gfw] Create metadata table to record e.g. latest month and year when change detection was ran by GFW


rudokemper:
## Feature Request

Currently, the `globalforestwatch` connector downloads and writes available alert data from Global Forest Watch, but unlike the `alerts_gcs` script, does not handle any statistical metadata. 

This is a problem when no alerts are detected for a number of months. There might be a scheduled Windmill script running to check for new alerts, but if we don't download any, then on the GC Explorer front end, the "Alert detection range" will show the date of the latest alert, not when the algorithm last ran.

An example might be helpful: let's say we have a script to fetch VIIRS fire detection data in Springfield, VA. VIIRS data is published daily, and we run the script daily. But if the last fire detected was in March 2025, then our front end will show the following:

> Alert detection range: 01-2025 to 03-2025
> Date of most recent alerts published: 03-2025

As there is no `__metadata` table to read to provide a more accurate record of the alert detection range.

Now, it is not known to me if an equivalent to `alerts_history.csv` exists for the GFW API. That is, for each alert that we support, does the GFW API provide an endpoint which returns the date range of the dataset?

If so, we can use that to construct a `__metadata` table, similar to `alerts_gcs`.

If not, then we can do some research on the cadence at which each dataset is updated (for example, for GLAD it is every 14 days if I'm not mistaken), and use that information to construct dates during script runtime.

IamJeffG:
Great points. Let's add the postgres table.

PR #156, opened 2025-10-02
Construct and write a `__metadata` table for GFW alerts

rudokemper:
## Goal

Closes https://github.com/ConservationMetrics/gc-scripts-hub/issues/145.

## What I changed

* I did some research and learned that each of the GFW alert sources that we support are updated daily.
* I added `prepare_gfw_metadata()` which uses `datetime` to determine a month and year value based on `datetime.now()` (given the previous bullet), and prepare the rest of the metadata records. These are then written to the DB per a metadata table schema provided by `create_gfw_metadata_table()`, similar to how we do this in `alerts_gcs`.
* The e2e test now checks for the `__metadata` table. I also added a different test to specifically check that months where 0 alerts are detected are written as desired to the database, as that was the original problem that led to https://github.com/ConservationMetrics/gc-scripts-hub/issues/145 being filed. Tests mock `datetime` to be deterministic.
* I added a minor fix to first check if there are any alerts before trying to write to DB.

## What I'm not doing here

* There is an edge case where the script fetches 0 alerts and therefore not create an alerts table, but will still create and write to a `__metadata` table reporting 0 detections per each month in the detection range. In that case, it's not possible to set up an alerts dashboard as there is no alerts table to work with, and the `__metadata` table is kind of vestigial. But I'm not sure if this is a problem. It might still be interesting to have the `__metadata` table to at least use as a reference to show that no alerts were detected for the given params.

### DIFF
--- a/f/connectors/globalforestwatch/gfw_alerts.py
+++ b/f/connectors/globalforestwatch/gfw_alerts.py
@@ -272,7 +272,7 @@ def prepare_gfw_metadata(alerts: list, type_of_alert: str, minimum_date: str):
     # Count alerts by day
     alerts_by_day = {}
     for alert in alerts:
-        # Extract date from alert (format varies by alert type)
+        # Extract date from alert (field name is different for VIIRS fire alerts)
         if type_of_alert == "nasa_viirs_fire_alerts":
             date_str = alert.get("alert__date")
         else:


f/connectors/globalforestwatch/gfw_alerts.py
```diff
@@ -193,3 +219,129 @@ def format_alerts_as_geojson(alerts: list, type_of_alert: str):
...
+
+
+def create_gfw_metadata_table(cursor, table_name):
+    """Create the metadata table for GFW alerts."""
+    cursor.execute(
+        sql.SQL("""
+        CREATE TABLE IF NOT EXISTS {metadata_table} (
+            _id character varying(100) NOT NULL PRIMARY KEY,
+            month bigint,
+            year bigint,
```
IamJeffG, line 340: In your PR description you say that these are updated _daily_ but we are only tracking updates at a monthly resolution? Why?

IamJeffG: Also: there's really no need for `month` and `year` to be `bigint`. `SMALLINT` would be fine for these types of vals.

f/connectors/globalforestwatch/gfw_alerts.py
```diff
@@ -193,3 +219,129 @@ def format_alerts_as_geojson(alerts: list, type_of_alert: str):
...
+        "radd_alerts": "deforestation",
+        "modis_alerts": "deforestation",
+    }
+
+    description_alerts = alert_descriptions.get(type_of_alert, "deforestation")
+
+    # Count alerts by month
+    alerts_by_month = {}
+    for alert in alerts:
+        # Extract date from alert (format varies by alert type)
```
IamJeffG, line 273: I am not seeing where we handle any format other than `YYYY-MM`. Please explain.

f/connectors/globalforestwatch/gfw_alerts.py
```diff
@@ -193,3 +219,129 @@ def format_alerts_as_geojson(alerts: list, type_of_alert: str):
...
+        if type_of_alert == "nasa_viirs_fire_alerts":
+            date_str = alert.get("alert__date")
+        else:
+            date_str = alert.get(f"{type_of_alert}__date")
+
+        if date_str:
```
IamJeffG, line 279: I am on the fence about whether to suggest we use the `datetime` module for the parsing and range comparison. Normally that would be the best practice in part because of all the possible edge cases it covers.

Like, I was thinking the comparison in particular could be simplified down to 1 line:
```
   if start_date <= alert_date < end_date:
```

But in this case, your parsing and range comparison is so simple that let's keep as-is. And it's the parsing side of things that might start to get more complicated by coercing everything into a proper `datetime`.

IamJeffG (approved):

Issue #168, opened 2025-10-21
[connectors/odk] configuration parameter rendered as string


nicopace:
The parameter that requests for url, user, password and project id:
https://github.com/ConservationMetrics/gc-scripts-hub/blob/89f90ebdd4ce5d4aa3f030a8e2df97c6cf39e499/f/connectors/odk/odk_responses.py#L61

Is being rendered as a text input:
<img width="732" height="211" alt="Image" src="https://github.com/user-attachments/assets/b8fbd6f6-6eb0-4ae2-814c-9d6f44857f9d" />

Shouldn't it be rendered differently?

rudokemper:
Yeah this is a bug having to do with the switch over to the Windmill Hub ODK resource type. I will fix this ASAP since we have a new partner using ODK.

PR #169, opened 2025-10-21
fix: set correct odk resource type

rudokemper:
## Goal

Closes https://github.com/ConservationMetrics/gc-scripts-hub/issues/168.


nicopace:
Screenshot 2025-10-21 at 12 12 47 PM

Amazing... thanks @rudokemper !

PR #170, opened 2025-10-22
Add documentation about user roles in Windmill

rudokemper:
## Goal

Closes https://github.com/ConservationMetrics/gc-programs/issues/35.

## What I changed

Added a section to the README about user roles and permissions, and how we recommend using Operator.

## What I'm not doing here

Documentation in `gc-deploy` is forthcoming shortly.

## LLM use disclosure

None


rudokemper:
Thanks. I clarified the process a bit and added screenshots.

Issue #101, opened 2025-05-22
[comapeo] Support downloading tracks in addition to observations


rudokemper:
## Feature Request

https://github.com/digidem/comapeo-cloud does not yet support returning tracks, but when it does -- or, when we know what the endpoint route will be and data schema will look like, and maybe have a fixture to work with -- we should expand the CoMapeo fetch scripts to also download tracks in addition to observations.

rudokemper:
New routes (to be shipped soon):

- `/projects/:projectPublicId/observation` (remove `s`)
- `/projects/:projectPublicId/track`
- `/projects/:projectPublicId/preset`
  - This returns fieldRefs and iconRef
- `/projects/:projectPublicId/icon/:docId`

PR #180, opened 2025-11-11
In `comapeo_pull`, support downloading tracks in addition to observations

rudokemper:
## Goal

Closes https://github.com/ConservationMetrics/gc-scripts-hub/issues/101.

## What I changed and why

* Added `SAMPLE_TRACK` fixture (based on a real response from Comapeo API) and used this to drive TDD.
* Changed script name to `comapeo_pull` since it's fetching more than just observations.
* Added `transform_comapeo_tracks()` to convert track locations into GeoJSON LineString geometry with a parallel timestamps array stored in properties. 
* Added `download_project_tracks()` and refactored observation fetching into `_fetch_comapeo_data()` helper to reduce duplication.
* Similarly extracted `_apply_preset_data()` and `_add_project_metadata()` helpers to reduce duplication between observations and tracks transformation logic.
* Updated data storage to create separate `_observations` and `_tracks` tables per project.

See https://github.com/ConservationMetrics/gc-scripts-hub/issues/101#issuecomment-3498032574 for additional rationale about these decisions.

## LLM use disclosure

I used Claude Sonnet 4.5 to help generate tests.

### DIFF
--- a/f/connectors/comapeo/README.md
+++ b/f/connectors/comapeo/README.md
@@ -78,16 +78,22 @@ This endpoint retrieves the binary data of a specific attachment, such as a phot
 {
   "name": "My Preset",
   "terms": ["term1", "term2"],
-  "color": "#000000"
+  "color": "#000000",
+  "iconRef": {
+    "docId": "icon_doc_id",
+    "url": "https://comapeo.example/projects/abc123/icon/icon_doc_id"
+  }
   ...
 }
 ```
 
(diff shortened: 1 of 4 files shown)


rudokemper:
Interesting to consider. To be honest, as this codebase has expanded, I think I've kept some of the legacy patterns from the very first Kobo pull script from frizzle v1 without really taking the time to question them. They work fine (proven to be robust in production for months now) and I've always made an effort to abstract away anything that is duplicative into modules, but your point is well taken about the pileup of custom code for similar patterns in many places. I think your observation could probably apply to many of the scripts that fetch data+attachments from an API. 

I think I will go ahead and merge this as I don't expect that the CoMapeo API will change much from here on out. But I appreciate you bringing this up.

PR #188, opened 2025-12-16
Save a GeoJSON file with missing attachments in CoMapeo

rudokemper:
## Goal

To write a GeoJSON file with `_missing_attachments` suffix in the filename, which a Guardian Connector user can access via Filebrowser to track down missing attachments in CoMapeo.

## Screenshots

image

RuntimeError showing number of attachments and observations affected in the logs.

## What I changed and why

* Store error info (URLs and messages) mapped to observation IDs
* Added `save_missing_attachments_geojson` helper fn to save a GeoJSON file of the observations with missing attachments, with `attachment_download_url` and `attachment_download_error` added as props.
* Update the final RuntimeError to include number of attachments that failed to download, and number of observations affected

## What I'm not doing here

Streaming observations to a file. I thought that would be good as a separate refactor that can be applied in a bunch of places.

## LLM use disclosure

Tests with Cursor

### DIFF
--- a/f/connectors/comapeo/tests/conftest.py
+++ b/f/connectors/comapeo/tests/conftest.py
@@ -1,5 +1,6 @@
 import re
 from dataclasses import dataclass
+from typing import Optional
 
 import pytest
 import responses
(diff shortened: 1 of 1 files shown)


f/connectors/comapeo/tests/conftest.py
```diff
@@ -160,6 +160,86 @@ class CoMapeoServer:
     )
 
 
+@pytest.fixture
+def comapeoserver_with_failing_attachments(mocked_responses):
+    """A mock CoMapeo Server with observations that have failing attachments"""
+
+    @dataclass
+    class CoMapeoServer:
```
IamJeffG: what's this type represent?  It feel its name is probably misleading -- and also overloaded (compare to `comapeo_server(TypedDict)` in `f/connectors/comapeo/comapeo_pull.py`)

IamJeffG (approved):

Issue #191, opened 2025-12-17
[Spike] Is it possible to add a connector for SMART?


rudokemper:
## Feature Request

SMART is a complex suite of tools. There is a mobile app [SMART Mobile](https://smartconservationtools.org/Download/SMART-Mobile), which interacts with either [SMART Desktop](https://smartconservationtools.org/Download/SMART-Desktop) or a web server [SMART Connect](https://smartconservationtools.org/en-us/Download/SMART-Connect).

Somewhere in the mix, there is also a variant of SMART Mobile called [SMART Collect](https://impactsmart.azurewebsites.net/Download/SMART-Collect) for "citizen science in remote areas", and then there is [CyberTracker](https://cybertracker.org/) which is often linked to in the SMART documentation, and seems to be a different suite of tools that is integrated with SMART. It's all a bit confusing.

It could be good for us to support SMART, even if no partner is currently asking for it.

I would like to try to see what it takes to get some SMART data into Guardian Connector. My best guess is...

1. In either SMART Desktop or SMART Connect, set up a project.
2. Invite a SMART Mobile device to that project.
3. Collect some data, "sync" it with SMART Desktop / Connect.
4. Then there must be some way to export the data.

I tried the Desktop route first, but it's very buggy on my Pop_OS! operating system. Lots of errors thrown everywhere. The release notes do say "The recommended Linux OS (that we used to build and test it) is Ubuntu 22."

For SMART Connect, they note...

> We recommend users adopt the SMART Partnership supported SMART Connect as a Service (SCaaS) option. To place a request for a SCaaS server [here](https://smartconservationtools.org/Download/SMART-Connect)
> If you are interested in deploying your own SMART Connect on-premise server, please contact us at info@smartconservationtools.org for guidance and the script.

I did write to them to see if we can get access. I reckon spinning up our own server to test with is the best bet for exploration.

PR #192, opened 2025-12-18
Add a SMART connector script

rudokemper:
## Goal

Towards (but does not close) https://github.com/ConservationMetrics/gc-scripts-hub/issues/191. Another PR forthcoming.

## What I changed and why

* Created a real export fixture from SMART Desktop + Mobile.
* Added a `smart_patrols.py` script that takes a SMART patrol export XML file and parses it to make a clean dataset of observation records in GeoJSON format.
* Leverage the usual pattern for geospatial data of saving the original (XML) file, saving a GeoJSON-transformed version, and calling `save_geojson_to_postgres` to save to db.

Should be straightforward for review.

## What I'm not doing here

Adapting the Dataset Uploader App to allow you to upload this file. That is coming in a next PR.

## LLM use disclosure

Supervised test generation with cursor.

### DIFF
--- a/f/connectors/smart/README.md
+++ b/f/connectors/smart/README.md
@@ -22,3 +22,22 @@ The XML file is also saved to the project folder for reference and can be used t
 SMART Connect allows SMART users to store and manage data on a web connected database.
 
 **TODO: Add SMART Connect scripts**: see [`gc-programs/smart/smart-connect`](https://github.com/ConservationMetrics/gc-programs/tree/main/smart/smart-connect) for more information on how to set up a SMART Connect server.
+
+
+## 📚 Reference
+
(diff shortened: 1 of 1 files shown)


IamJeffG (approved):
my only question is what sort of heterogeneity might appear in terms of the schema (mostly the set of fields) of smart XML files? And is this script special-cased to the one example we have?

rudokemper:
Pretty hard to say, honestly. I did export three XMLs to be sure, and all of these fields appeared in them (they are required fields when "starting a patrol"). But SMART is a complex toolchain and it would not surprise me if there are divergences. That said, there is NO documentation about their data structures so it's tricky to know without more samples. I vote that we deal with that when the script gets some usage from XML exports generated in the wild.

Issue #203, opened 2026-02-02
[alerts_gcs] Handle case where a file is updated upstream (current behavior: ignore)


rudokemper:
## Current Buggy or Suspicious Behavior

We have encountered before an issue where a change detection alert GeoJSON is updated upstream, and we need to update the database and files. This resulted in a fix where the `alerts_gcs` script, in `sync_gcs_to_local()` [does a check](https://github.com/ConservationMetrics/gc-scripts-hub/blob/f8a976e84e13003fffdbb665f7ac46b4dbde69ed/f/connectors/alerts/alerts_gcs.py#L362-L378) to see if the upstream file is up to date with the local file. 

It appears another adjustment is needed, likely in the DB Writer, as we have observed that the database records are not being updated with the change, but the actual GeoJSON files written to file storage are.

PR #204, opened 2026-02-03
Add test case for update logic in `alerts_gcs`

rudokemper:
## Goal

Closes https://github.com/ConservationMetrics/gc-scripts-hub/issues/203.


nicopace (approved):
Looks good to me Rudo.

PR #211, opened 2026-02-21
Regenerate metadata / lockfiles, and fix transitive deps

rudokemper:
## Goal

To bump dependency requirements and apply YAML formatting fixes generated by `wmill [script/flow/app] generate-metadata`, along with a small number of additional manual fixes by me.

> [!NOTE]
> I made sure to test these changes in runtime as well as `tox`. I tested connector scripts, the dataset uploader app, and our alerts flow: everything works as expected.

## Context

The reason I went ahead and did this is that after the recent psycopg3 upgrade, most scripts stopped working due to incorrect dependencies. For example, this was the result of a Kobo fetch script run:

```
  File "/tmp/windmill/wk-default-ff7b462c1d36-E1aQP/019c7dd0-fb93-6d9f-5388-ce0a736854c2/wrapper.py", line 10, in <module>
    from f.connectors.kobotoolbox import kobotoolbox_responses as inner_script
  File "/tmp/windmill/wk-default-ff7b462c1d36-E1aQP/019c7dd0-fb93-6d9f-5388-ce0a736854c2/f/connectors/kobotoolbox/kobotoolbox_responses.py", line 14, in <module>
    from f.common_logic.db_operations import StructuredDBWriter, conninfo, postgresql
  File "/tmp/windmill/wk-default-ff7b462c1d36-E1aQP/019c7dd0-fb93-6d9f-5388-ce0a736854c2/tmp/f/common_logic/db_operations.py", line 12, in <module>
    from psycopg import Error, connect, errors, sql
  File "/tmp/windmill/cache/python_3_11/psycopg[binary]==3.2.3/psycopg/__init__.py", line 9, in <module>
    from . import pq  # noqa: F401 import early to stabilize side effects
    ^^^^^^^^^^^^^^^^
  File "/tmp/windmill/cache/python_3_11/psycopg[binary]==3.2.3/psycopg/pq/__init__.py", line 18, in <module>
    from . import abc
  File "/tmp/windmill/cache/python_3_11/psycopg[binary]==3.2.3/psycopg/pq/abc.py", line 13, in <module>
    from .._compat import Self, TypeAlias
  File "/tmp/windmill/cache/python_3_11/psycopg[binary]==3.2.3/psycopg/_compat.py", line 46, in <module>
    from typing_extensions import TypeVar
ModuleNotFoundError: No module named 'typing_extensions'
```

I remembered that at some point, I began generating and maintaining lock files on a per-script basis because certain scripts would break when running `generate-metadata`, due to transitive dependencies that are not captured correctly during metadata generation. I went ahead and fixed those as well by adding explicit requirement comments in the Python code, so we can rely on `generate-metadata` and eliminate the need for manual lockfile updates.

## What I changed and why

* Mass application of changes introduced by `generate-metadata` across all Windmill scripts, flows, apps.
* Most of the changes in this PR are boilerplate but the following are worth noting:
  * `f/common_logic/data_conversion.py`
    * had a lazy import of `lxml` (via `f.connectors.smart.smart_patrols`) that Windmill's metadata generator couldn't detect, causing it to be removed from the lockfile. Solution was to add a requirement comment at the top of the file.
    * similarly added `xlrd` to requirements in `data_conversion.py` to support legacy Excel files.
  *  `f/connectors/odk/tests/conftest.py`
    * Upgrading `urllib3` (1.26.20 -> 2.5.3) enforces Content-Length headers strictly. This caused an ODK test to fail because test fixture had a hard-coded `Content-Length: 3632` header for some reason, but the actual test file was only 920 bytes. So I removed that hard-coded header.
    * FWIW, I remembered that the ODK script used to have gnarly requirement incompatibilities due to mismatches between its imports and pinned dependencies, which was a motivation for starting to maintain per-script requirements. But that doesn’t seem to be an issue anymore, possibly due to recent updates to [`pyodk`](https://github.com/getodk/pyodk). I was expecting to wrestle with this again, but no problems this time around.
  * `f/apps/gc_dataset_importer.app`
    * The `generate-metadata` tool itself seems to have set everything to `py: 3.11` which is fine as that is what we decided to stick with for now. Not sure why it was set to `py: 3.13` to begin with; maybe an artifact of the app builder being used in an instance with that version of Python set.

> [!NOTE]
> 
> This PR will muddy up the commit history, so I propose to add the commit to `.git-blame-ignore-revs` after merging.

## LLM use disclosure

None

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,6 +101,7 @@ To maintain consistency across the repository, please follow these conventions w
 #### Language and Dependencies
 * While Windmill supports multiple languages (Python, TypeScript, Go, etc.), for ETL-related workflows, use Python to maintain consistency across the repository.
 * Use Python 3.11 when setting dependencies in `«name».script.lock` and `requirements-test.txt` files, as it is the latest stable version we use for our Windmill instances and in CircleCI.
+* Generate lockfiles by either running a script in Windmill runtime (Windmill will create the lockfile for you) or by running the `wmill script generate-metadata` command. If you are having issues with transitive dependencies not included in the lockfile, add them to the top-of-script `# requirements` block.
 
 #### Script Metadata
 * Add a clear summary for each script in the `«name».script.yaml` file.


IamJeffG (approved):
Am I right that the advice going forward is that we don't need to write lock files manually?

More concretely, my experience with windmill is:
- running the script in windmill runtime, if no lockfile existed, it will create a good one for you.
- You can also run `generate-metadata` on the CLI without needing to sync or run the script on a deployment

I could imagine a world in which surgical tweaks to the lockfile (that generate-metadata created) are needed, but the problems to-date don't sound like that is the correct.  I'd much prefer handling edge cases like that in the top-of-script `# requirements` block.



> This PR will muddy up the commit history, so I propose to add the commit to .git-blame-ignore-revs after merging.

No objection, but I don't think it's that important either to do this. The only lines changes are "uninteresting" ones.

PR #226, opened 2026-03-18
Add a Windmill app to apply Local Contexts labels to a dataset

rudokemper:
## Goal

Closes https://github.com/ConservationMetrics/gc-scripts-hub/issues/225.

### Step 1: select dataset and label set to apply

image

### Step 2: select labels to annotate

image

### Step 3: Save and finish

image

_fix: I am hiding that "Ready to write?" label now, if upload successful_

## What I changed and why

- See `## What this app does, code-wise` in the application `README.md` for a description of how this app works. Otherwise I recommend test-driving the app on `demo`, per `## How to review` below.
- The only "common logic" code I had to add was a `truncate_table()` fn to truncate a table using psycopg in `db_operations.py`. I am pleased that all of the existing code e.g. to check existence of a table, write to table, etc. worked just fine in this application!
- Per feedback from LC, I capitalized "Project" and "Label" everywhere.
- Also per feedback from LC, I am utilizing the Open to Collaboration endpoint to ensure only Community API keys can fetch data.

## What I'm not doing here

Once, we considered an additional step for this application where IF a LC label set does not exist, then run the script to fetch and write a table, and proceed from there. For now I am treating the run or schedule of the "Local Contexts: Fetch Labels" script as a separate operation.

## How to review

Per the app `README.md`,

> The `app.yaml` file generated by Windmill is highly impractical to review directly. Instead, it is recommended to sync the app to a Windmill instance, and review the code in the Windmill App builder UI.

This is absolutely the case for this PR. Also, +1299 out of the +1702 line changes are in `app.yaml`! The only **code** I would actually review are:

- `1_fetch_lc_labels_and_existing_mappings.inline_script.py` (+140)
- `3_write_label_mapping_to_database_table.inline_script.py` (+85)
- `db_operations.py` (+53)
 
Otherwise, I recommend test-driving the app on `demo`. I am happy to jump on a call to walk through how this works, or discuss feedback instead of through comments or code review here on Github!

## LLM use disclosure

None


nicopace (approved):
aside from the mentioned comments, it looks good!

PR #235, opened 2026-05-03
Add an import script for CyberTracker

### LINKED ISSUE #233: Add a connector for CyberTracker
## Feature Request

CyberTracker is commonly used across Southern Africa. It would be good to first understand how people are using it, and whether they are leveraging the mobile app only or also the online platform. For this issue, let's do the research first before adding any code.

rudokemper:
## Goal

Towards https://github.com/ConservationMetrics/gc-scripts-hub/issues/233.

Adds a connector import script for CyberTracker backups. See the README for more information on why I chose to use this particular format, and how to generate a backup.

> [!NOTE]
> +217 lines changed is just the fixture, and +80 lines are boilerplate like yaml. That still leaves +508 lines. However, the code largely follows a known ingest pattern shared across connector scripts, so I hope this isn't off-putting to the reviewer.

## ~Screenshots~ How I know this works

I tried it in runtime with several CyberTracker backups.

## What I changed and why

* Added a `0.json` fixture which is what holds the CyberTracker observation data from the Backup. This fixture is composed of data from multiple project Backups, and I made an effort by checking it against several other Backups to ensure we don't special-case the connector code.
* Added a `README.md` that comprehensively documents my learnings, and how to get / use this Backup.
* Added `cybertracker_observations_from_backup.py` which holds the core logic. 
  * Some patterns are followed from other connectors (i.e. SMART) like having modular `transform_cybertracker_data` and `parse_cybertracker_data` functions that can be imported elsewhere, like for the Dataset Uploader. (See #236 for that.)
  * The main thing this script is doing is pulling out observations data that has X and Y coordinates. This is the main data type of interest. CyberTracker also records tracks in the background and stores these as KMZ. I opted not to work with those for now (and clarified that decision in the readme and code) -- as these are just tracks without any annotation of interest. They can also be uploaded directly from the Backup archive.
  * There is a good amount of code here that uses heuristics to detect fields. I couldn't think of a better way around this given the messy structure of the upstream data. But ideas and pushback welcome around any of the decisions made in the code. 

Last thing I'll add is that we know there are users in the field with real CyberTracker data on their devices, and this Backup is basically the only way to get the data out aside from uploading to a server (like CyberTracker Online, Earth Ranger) that they seemingly do not have access to! So while it may seem unwieldy to work with data that is structured to be able to import back into CyberTracker later, it appears to be the only way. (To be explicit: CyberTracker mobile does not have an "Export as CSV" option.)

## What I'm not doing here

Supporting uploading via the dataset uploader. That is added by https://github.com/ConservationMetrics/gc-scripts-hub/pull/236.

## LLM use disclosure

I used Claude Opus 4.7 in Cursor to inspect CyberTracker outputs and generate the parsing code and field heuristics, which I then edited slightly.


f/connectors/cybertracker/cybertracker_observations_from_backup.py
```diff
@@ -0,0 +1,293 @@
...
+        and "/" not in raw_key
+        and value
+        and all(_looks_like_uid(v) for v in value)
+    ):
+        # Parent repeat fields contain child uids; prefer the child rows which
+        # include the actual filenames under a ".../..." key.
+        return None
+
+    if isinstance(value, list):
+        return value[0] if len(value) == 1 else value
```
IamJeffG: will this be surprising to callers?

What if 2 different records are processed: record A has a 1-item list, record B has a 2-item list: will anything break because record A (wrongly?) had its type inferred as not-a-list?

f/connectors/cybertracker/tests/cybertracker_observations_from_backup_test.py
```diff
@@ -0,0 +1,110 @@
...
+    assert props["_end"] == "2026-04-30T11:41:02.531-04:00"
+    assert props["_username"] == "field_team_alpha"
+
+    assert props["number_of_species"] == 7
+    assert props["audio_recording"] == "e234a9b27a4a494a9800a918b1d034ce.wav"
+    assert props["photo_of_site"] == "PHOTO_20260430_114040.jpg"
+    assert props["collect_data"] == [
+        "s20260430153623050_collect_data/i20260430153647038_ceiba_canopy",
+        "s20260430153623050_collect_data/i20260430153657095_fire_scar_history",
+    ]
```
IamJeffG: This mismatch is similar to what I am talking about in my comment for `_normalize_field_value`. 

It seems to me life would be simpler if something that might be list is always a list.

Thoughts?

f/connectors/cybertracker/cybertracker_observations_from_backup.py
```diff
@@ -0,0 +1,293 @@
...
+            delete_geojson_file=False,
+        )
+        logger.info(
+            f"CyberTracker observations successfully written to database table: [{db_table_name}]"
+        )
+    else:
+        logger.warning("No observations found in CyberTracker JSON")
+
+
+def transform_cybertracker_data(data: dict, dataset_name: str = None) -> dict:
```
IamJeffG, line 101:
```suggestion
def transform_cybertracker_data(data: dict, dataset_name: Optional[str] = None) -> dict:
```

IamJeffG: I see now that this is ignored.  I am not seeing what it's "other transformation functions" we are being consistent with. Tempted to drop this optional param.

f/connectors/cybertracker/cybertracker_observations_from_backup.py
```diff
@@ -0,0 +1,293 @@
...
+    logger.info(f"Extracted {num_observations} observations from CyberTracker JSON")
+
+    # Create project directory in datalake
+    project_dir = Path(attachment_root) / db_table_name
+    project_dir.mkdir(parents=True, exist_ok=True)
+
+    # Save raw JSON file to datalake
+    json_save_path = project_dir / json_path.name
+    json_save_path.write_text(
+        json_path.read_text(encoding="utf-8-sig"), encoding="utf-8"
```
IamJeffG, line 66: probably worth a code comment that we're not making an archival copy but copying nad changing the text encoding.

IamJeffG (approved):

f/connectors/cybertracker/cybertracker_observations_from_backup.py
rudokemper: Let me fix this in a follow-on PR.

PR #236, opened 2026-05-03
Support CyberTracker uploads in GC Dataset Uploader

### LINKED ISSUE #233: Add a connector for CyberTracker

rudokemper:
## Goal

Stacked onto https://github.com/ConservationMetrics/gc-scripts-hub/pull/235; closes #233 once both merged.

## ~Screenshots~ How I know this works

I tried it in runtime with several CyberTracker backups.

## What I changed and why

* Most of the work happens in the data detection and conversion code. The patterns here are similar to what we did for SMART.
  * `detect_structured_data_type()` does an additional check for JSON to sniff if it is a CyberTracker file, based on a set of known keys that appear in every session / observation. 
    * On the one hand, I don't love the approach used here as the CyberTracker keys are not super deterministic. Like, to compare with what we do SMART, checking for a set of keys like `schemaHash`, `records`, and `db` is not the same as sniffing for an explicit declaration of a SMART XML namespace). 
    * On the other hand, I can't imagine any of our users coming to Guardian Connector with a wild JSON file with this specific set of keys that isn't from CyberTracker. (To be honest, I don't imagine many users trying to upload a JSON file that isn't GeoJSON at all.) 
    * So I find this to be an acceptable solution. The app will fail loudly if something goes wrong in the conversion, anyway, so we at least don't risk writing non-CyberTracker data to the warehouse.
  * `convert_data()` now accepts `cybertracker` file format; `read_cybertracker()` calls `parse_cybertracker_json()` added by #235.
* The app already accepted JSON files, so the detection / conversion changes will handle the conversion to GeoJSON. The only additional thing that happens in the app is that we auto-set "CyberTracker" as data source, as we will already know this from importing. And call `transform_cybertracker_data` to add the `data_source` col.

## LLM use disclosure

None

### DIFF
--- a/f/apps/gc_dataset_importer.app/2_upload_and_convert_file.inline_script.py
+++ b/f/apps/gc_dataset_importer.app/2_upload_and_convert_file.inline_script.py
@@ -167,6 +167,7 @@ def main(uploaded_file, dataset_name, table_exists, table_name, db: postgresql):
             None,
             output_filename,
             output_format,
+            file_format,
             new_rows,
             updates,
             new_columns,
@@ -176,4 +177,4 @@ def main(uploaded_file, dataset_name, table_exists, table_name, db: postgresql):
     except Exception as e:
         error_msg = f"Error during file upload and conversion: {e}"
         logger.error(error_msg)
-        return False, error_msg, None, None, None, None, None, None
+        return False, error_msg, None, None, None, None, None, None, None
--- a/f/common_logic/data_conversion.py
+++ b/f/common_logic/data_conversion.py
@@ -871,7 +871,7 @@ def read_cybertracker(path: Path):
     dict
         GeoJSON FeatureCollection with observation features.
     """
-    # Lazy import to avoid forcing CyberTracker dependencies on all users of data_conversion.py
+    # Lazy import to avoid forcing this script dependencies on all users of data_conversion.py
     from f.connectors.cybertracker.cybertracker_observations_from_backup import (
         parse_cybertracker_json,
     )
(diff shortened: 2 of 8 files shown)


f/common_logic/data_conversion.py
```diff
@@ -828,6 +856,29 @@ def read_kml(path: Path):
...
+    ----------
+    path : Path
+        Path to the CyberTracker JSON file.
+
+    Returns
+    -------
+    dict
+        GeoJSON FeatureCollection with observation features.
+    """
+    # Lazy import to avoid forcing CyberTracker dependencies on all users of data_conversion.py
```
IamJeffG: CyberTracker dependencies make be think of their 3rd party SDK but I think you just mean "our gc-scripts-hub parsing code"


```suggestion
    # Lazy import to avoid forcing this script dependencies on all users of data_conversion.py
```

That said, the deps of that script are light. This could probably go at top-level if you want. Either is fine by me.

f/apps/gc_dataset_importer.app/2_upload_and_convert_file.inline_script.py
```diff
@@ -176,4 +177,4 @@ def main(uploaded_file, dataset_name, table_exists, table_name, db: postgresql):
     except Exception as e:
         error_msg = f"Error during file upload and conversion: {e}"
         logger.error(error_msg)
-        return False, error_msg, None, None, None, None, None, None
```
IamJeffG, line 179: Do Windmill apps support passing around something like a https://docs.python.org/3/library/dataclasses.html or https://docs.python.org/3/library/collections.html#collections.namedtuple ?

Big lists of anonymous params like this are subject to bugs and give me the heebie-jeebies ;-)

IamJeffG (approved):

PR #241, opened 2026-05-19
Add an EpiCollect connector

### LINKED ISSUE #240: Add an EpiCollect connector
## Feature Request

A prospective partner is using this tool for data collection. It is a different survey builder solution than the XLSForm apps (Kobo and ODK). But it has some of the interesting features that our partners like to use such as Location, Photo, Audio, Video, along with typical survey options.

There is an API: https://developers.epicollect.net/

rudokemper:
## Goal

Closes https://github.com/ConservationMetrics/gc-scripts-hub/issues/240. Adds a connector for the surprisingly pleasant-to-use [EpiCollector](https://epicollect.net/) mobile application.

## How I know this works

I tested it in production with 4 different EpiCollector projects -- two created by me, and two that are public demos.

## Notes to the reviewer

I would put primary emphasis on reviewing `epicollect_pull.py` (+398 lines) and in particular the aspects that deviate from the typical connector pattern, mentioned in the next section.

## What I changed and why

- Server responses for submissions in `server_responses.py` reflect most of the interesting different field types possible in EpiCollect. (That is, not _every field under the sun_ as many just return a String.)
- In `epicollect_pull.py`, things of note **(and maybe to emphasize for review as the rest follows similar patterns**) include:
  - API auth is optional, as public EpiCollect5 projects work without credentials. Public projects also return full media URLs as field values rather than bare filenames, so these are normalized before DB insertion.
  - Media fields (photo/audio/video) are detected ahead of time by cross-referencing the form's inputs with the default project mapping, since EpiCollect doesn't wrap attachments in an envelope like Kobo/ODK do.
  - A 1-second delay between paginated requests as a courtesy to the shared public API.
  - We download a project logo thumbnail to {attachment_root}/{table_name}/logo.jpg as a separate step before any entries are processed. 
- As usual, added tests, Windmill script YAML, etc.

## What I'm not doing here

- Because API keys are only generated per-project, I didn't think it made sense to make a Windmill Resource for EpiCollect -- there is no reusability value in doing so, and just adds more busy work in creating a resource.

## LLM use disclosure

Because this was a very straightforward connector (tabular data provided by API), I generated most of this through careful prompt engineering in Cursor, and then made fixes primarily to documentation and tests.


rudokemper:
@IamJeffG ping

f/connectors/epicollect/epicollect_pull.py
```diff
@@ -0,0 +1,398 @@
...
+
+
+def main(
+    project_slug: str,
+    db: postgresql,
+    db_table_name: str,
+    client_id: int | None = None,
+    client_secret: str | None = None,
+    attachment_root: str = "/persistent-storage/datalake",
+):
```
IamJeffG, line 33:
```suggestion
):
    """
    API auth is optional, as public EpiCollect5 projects work without credentials.
    """
```

f/connectors/epicollect/epicollect_pull.py
```diff
@@ -0,0 +1,398 @@
...
+
+from f.common_logic.db_operations import StructuredDBWriter, conninfo, postgresql
+from f.common_logic.file_operations import save_data_to_file
+
+
+BASE_URL = "https://five.epicollect.net"
+_MEDIA_TYPES = frozenset({"photo", "audio", "video"})
+_MEDIA_FORMAT = {"photo": "entry_original", "audio": "audio", "video": "video"}
+_PAGE_SIZE = 250
+_PAGE_DELAY_S = 1.0
```
IamJeffG:
```suggestion
_PAGE_DELAY_S = 1.0  # between paginated requests; try to be a nice neighbor
```

IamJeffG (approved):

PR #243, opened 2026-06-08
Fix CI: work around tox-docker Gateway KeyError on newer Docker

### LINKED ISSUE #242: Tests are failing on `alerts`, likely a Python version issue in CI
[Tests are failing on alerts](https://app.circleci.com/pipelines/github/ConservationMetrics/gc-scripts-hub/786/workflows/0e4b0392-1b15-46ec-b97d-c936a249e409/jobs/786), unrelated to this PR. I suspect it was a byproduct of https://github.com/ConservationMetrics/gc-scripts-hub/commit/344c7e09e477ed76e350bf26e0d6817a415cef41. I believe I've seen this error before and we had to mess with Python versions in CI. Will file a separate issue.

_Originally posted by @rudokemper in https://github.com/ConservationMetrics/gc-scripts-hub/issues/241#issuecomment-4650170306_

rudokemper:
## Goal

Closes https://github.com/ConservationMetrics/gc-scripts-hub/issues/242. Unblock `alerts` tox env on CI. 

This was not a Python / pip version issue as initially suspected, but rather a discrepancy between `tox-docker` and `ubuntu-2404:current`.

## What I changed and why

We set `TOX_DOCKER_GATEWAY: localhost` in `.circleci/config.yml`. 

Newer Docker on `ubuntu-2404:current` no longer fills top-level `NetworkSettings.Gateway`, so tox-docker seems to crash with a KeyError. Claude Opus 4.7 did this diagnosis:

> tox-docker has a documented escape hatch for exactly this: if `TOX_DOCKER_GATEWAY` is set, it skips `container.attrs["NetworkSettings"]["Gateway"]` and uses `socket.gethostbyname(gateway)` instead. Setting it to localhost resolves to 127.0.0.1, which is what you want on `machine-executor` anyway (host networking).

- Trace lands in tox-docker, not our code; `pip freeze` is identical to previous green runs on CircleCI.
- `alerts` tests seems to be passing again in CI.

## What I'm not doing here

- Pinning `ubuntu-2404` to a dated tag.

## LLM use disclosure

Claude Opus 4.7 helped diagnose.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -50,3 +50,7 @@ jobs:
           command: tox -- -v
           environment:
             PIP_UPLOADED_PRIOR_TO: P7D
+            # Skip tox-docker's NetworkSettings.Gateway lookup (broken on newer
+            # Docker in ubuntu-2404:current). machine-executor reaches containers
+            # via localhost anyway.
+            TOX_DOCKER_GATEWAY: localhost
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -52,5 +52,6 @@ jobs:
             PIP_UPLOADED_PRIOR_TO: P7D
             # Skip tox-docker's NetworkSettings.Gateway lookup (broken on newer
             # Docker in ubuntu-2404:current). machine-executor reaches containers
-            # via localhost anyway.
+            # via localhost anyway. Workaround until https://github.com/tox-dev/tox-docker/pull/196
+            # is merged and released.
             TOX_DOCKER_GATEWAY: localhost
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -50,3 +50,7 @@ jobs:
           command: tox -- -v
           environment:
             PIP_UPLOADED_PRIOR_TO: P7D
+            # Skip tox-docker's NetworkSettings.Gateway lookup (broken on newer
+            # Docker in ubuntu-2404:current). machine-executor reaches containers
+            # via localhost anyway.
```
IamJeffG: would you mind labeling this as a temp workaround until https://github.com/tox-dev/tox-docker/pull/196 is released?

(if you agree that's true)

IamJeffG (approved):

PR #251, opened 2026-06-25
Improve save CSV to disk script, and apply for survey tools

rudokemper:
## Goal

One of the things we are discovering that people really like about Guardian Connector is the ability to use Filebrowser to download an ingested file version of a dataset. We currently do this for spatial datasets -- that is, when, say, the CoMapeo script is run, in addition to writing to Postgres it also saves a GeoJSON to disk. And our users have found value in then logging into Filebrowser to download the GeoJSON file for usage in QGIS or other third party tools.

Recently, a use case came up to also do the same for non-spatial datasets, like a KoboToolbox survey. One of our partners is building a survey with nested groups, and the way we handle those in the KoboToolbox script actually produces a more useful tabular dataset than what you get by downloading the data from KoboToolbox itself. So, we would like to make the CSV available on disk, much like we do for GeoJSON files.

In doing so, I thought it would also be good to harden CSV serialization so exported datasets preserve nested structures and handle missing values consistently.

## What I changed and why

1. Updates `save_data_to_file` so CSV exports can be generated directly from structured records. We now supports a list of dictionaries, builds a header row, keeps `_id` first when present, JSON-encodes nested lists/dicts, and writes missing/None values as empty cells so the CSV can round-trip cleanly back into Postgres.
2. Has all the survey tools (Kobo, ODK, EpiCollect) saving the file as CSV to disk.

- Justification for why this is right in **Goal**
- Trying it out in runtime (inspecting a few actual CSVs generated)
- Our whole test suite passing

## What I'm not doing here

I thought twice about whether `csv_to_postgres` should handle the saving to disk, so we don't need to do that in every connector. But I decided against it (for now) because `geojson_to_postgres` also doesn't handle this, so I preferred to be consistent.

## LLM use disclosure

I had Cursor generate the docstrings (which I then further simplified as they were overly verbose) and the tests were also heavily auto-complete written.

### DIFF
--- a/f/connectors/kobotoolbox/kobotoolbox_responses.py
+++ b/f/connectors/kobotoolbox/kobotoolbox_responses.py
@@ -347,6 +347,94 @@ def download_form_responses_and_attachments(
     return form_submissions, form_name, form_languages
 
 
+# TODO - YAGNI: Extract to f/common_logic/submission_flatten.py if/when ODK
+# flattening is required; not worth a common module until then.
+def _get_flattenable_kobo_rows(value):
+    """Return a Kobo field value as a list of rows to flatten, or ``None`` to leave it alone.
+
+    We only flatten values that look like a group of fields with slash-separated
+    keys. There are two such shapes:
+
+    A list of row dicts (repeat group / matrix), e.g.::
+
+        [
+            {"household_members/name": "Ada", ...},
+            {"household_members/name": "Bo",  ...},
+        ]
+
+    A single dict of slash keys (field-list group), e.g.::
+
+        {"location/lat": 1.0, "location/lng": 2.0}
+
+    Both are returned as a list of rows. Everything else (scalars, empty
+    containers, geolocation, attachments, and system lists/dicts like ``_tags``
+    or ``_validation_status``) returns ``None`` and is kept as-is.
+    """
+    # Every row must be a dict with at least one slash key. This excludes plain
+    # lists such as _tags, _attachments metadata, and empty repeats.
+    if (
+        isinstance(value, list)
+        and value
+        and all(isinstance(item, dict) for item in value)
+        and all(any(isinstance(k, str) and "/" in k for k in item) for item in value)
+    ):
+        return value
+
+    # A lone group dict has no repeat index, so wrap it as one row and let the
+    # same flatten loop handle it. Requiring "/" on every key excludes system
+    # dicts such as _validation_status.
+    if (
+        isinstance(value, dict)
+        and value
+        and all(isinstance(k, str) and "/" in k for k in value)
+    ):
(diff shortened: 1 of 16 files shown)


f/connectors/kobotoolbox/kobotoolbox_responses.py
```diff
@@ -56,13 +57,24 @@ def main(
...
+    save_csv_to_postgres(
+        db,
         db_table_name,
+        str(Path(db_table_name) / f"{db_table_name}.csv"),
+        attachment_root,
+        delete_csv_file=False,
+        id_column="_id",
         use_mapping_table=True,
         reverse_properties_separated_by="/",
     )
```
conservationtimothy: Question: What happens when a form has zero submissions? If I'm reading this right: `save_data_to_file` skips writing when data is empty (just logs a warning and returns), but we then call `save_csv_to_postgres` unconditionally, which tries to open the CSV that was never written, so a `FileNotFoundError`?

And there's a nastier variant I think: if a previous run did write the CSV and a later run returns zero entries, we'd skip the write but happily re-ingest the stale CSV from the last run and log "successfully written". Am I getting this right, or is there something upstream that guarantees we never reach here with empty data?

Previously `StructuredDBWriter.handle_output([])` was a graceful no-op, so this would be new behaviour.

rudokemper: Thanks for the feedback. You are right on both fronts! 

I decided to guard the connectors themselves so a zero-submission pull just logs and skips: https://github.com/ConservationMetrics/gc-scripts-hub/pull/251/commits/0fb4940f67a7ab84e0bfc872b4d27ecb3ea52dcc

Reasoning behind this concern decision: only the connector actually knows the pull came back empty, so a missing-file check in `csv_to_postgres` would stop the crash but not the case where we'd re-ingest last run's stale CSV and report success. That code's also shared with the upload app where the CSV always exists, so it didn't feel like the right home for this.

rudokemper: (Let me fix the failing tests)

rudokemper: The odk_responses CI failure was actually a test isolation issue, not a problem with the change itself. https://github.com/ConservationMetrics/gc-scripts-hub/pull/251/commits/f4612fbbb9eeefa21cb0f25a9a540625d39465ea

The change is technically out of scope, so I'd be fine to isolate it into its own PR.

conservationtimothy: Thanks! I am comfortable with keeping it in here.

conservationtimothy (approved):

PR #256, opened 2026-07-01
Flatten KoboToolbox matrix and repeat-group responses

### LINKED ISSUE #248: Flatten KoboToolbox matrix and repeat-group responses instead of dumping them as JSON blobs
## Feature Request

`kobotoolbox_responses.py` (and the ODK equivalent) passes Kobo API submissions through `transform_kobotoolbox_form_data()` with almost no structural normalization: geolocation gets special handling, but matrix questions and repeat groups do not.

When Kobo returns nested data (matrix rows, repeat groups, field-list groups), those values are typically **lists of dicts** with long slash-separated keys, e.g.:

```json
[
  {
    "Matrix_category/group_ux1bc08_person_1/group_ux1bc08_person_1_profession": "IT Professional",
    "Matrix_category/group_ux1bc08_person_1/group_ux1bc08_person_1_type_of_job: "Developer"
  }
]
```

Today that entire structure lands in **one cell**. Meaning that spreadsheet exports and SQL queries are unusable for matrix/repeat data, and users cannot analyze sub-questions without manual JSON parsing.

The existing `reverse_properties_separated_by="/"` logic only helps flat slash keys (e.g. `meta/instanceID` → `instanceID__meta`). It does not expand list-of-dict matrix payloads.

**Affected connector:** `f/connectors/kobotoolbox/kobotoolbox_responses.py`  
**Related plumbing:** `transform_kobotoolbox_form_data()`, `save_data_to_file()`, `csv_to_postgres`, form metadata in `{db_table_name}_metadata.json`

Let's consider ODK out of scope for this issue, but leave a `TODO` note for ODK or file a sister issue.

## Proposed solution

Add a `flatten_kobotoolbox_submission(submission, form_metadata)` step inside `transform_kobotoolbox_form_data()`.

1. Walk each submission key/value.
2. If the value is a `list[dict]` (repeat group / matrix payload from Kobo): drop the parent key, flatten each item into `{group}/{index}/{leaf_key}` with 1-based `{index}`. `{leaf_key}` = last path segment after `/` (or map via metadata when available).
3. If the value is a lone `dict` with slash keys (field-list group): same flattening, index `1`.
4. Optionally use `{db_table_name}_metadata.json` (`content.survey`) to recognize `begin_repeat` vs matrix/field-list groups and strip redundant internal group segments from column names — nice-to-have, not blocking v1.

The output should be one wide parent row per submission, no JSON blobs in matrix/repeat columns. Existing `reverse_properties_separated_by="/"` + `__columns` mapping keeps working on the flattened keys.

rudokemper:
## Goal

Closes https://github.com/ConservationMetrics/gc-scripts-hub/issues/248

**Note to the reviewer:** +340 of the line difference is just fixtures and `conftest` stuff. The actual code changes are +67 lines, and +90 are tests.

## What I changed and why

- Added `flatten_kobotoolbox_submission()` in the Kobo connector script and call it from `transform_kobotoolbox_form_data()` so repeat groups and matrix questions become wide `{group}/{index}/{leaf}` keys instead of one JSON cell. Existing `reverse_properties_separated_by="/"` then maps those to flat SQL columns. See README and docstrings for more info.

- In terms of the "why do this", by working with one of our highest tier partners who wanted to analyze matrix results and was not able to work with a JSON blob dumped in one cell. We had never encountered matrix questions in a Kobo questionnaire before this.
- In terms of "how I know this works", besides tests, I tried it in production with real data. The result is as described in the README. 

## What I'm not doing here

- Apply this to ODK (also uses XLSForm), but I did leave clear TODOs to track that future need if it arises.

## LLM use disclosure

As I was working with a partner in country and didn't have a lot of time, I used Claude Opus 4.7 in planning mode to scheme this work out, and had it implement it too. Then I made various edits to ensure the code and its documentation is legible to me. I screened the code carefully to make sure it is valid and maintainable, making a few improvements here and there.

### DIFF
--- a/f/connectors/kobotoolbox/kobotoolbox_responses.py
+++ b/f/connectors/kobotoolbox/kobotoolbox_responses.py
@@ -342,8 +342,8 @@ def download_form_responses_and_attachments(
     return form_submissions, form_name, form_languages
 
 
-# TODO - YAGNI: Extract to f/common_logic/submission_flatten.py if/when ODK repeat/matrix
-# flattening lands; not worth a common module until then.
+# TODO - YAGNI: Extract to f/common_logic/submission_flatten.py if/when ODK
+# flattening is required; not worth a common module until then.
 def _slash_keyed_rows(value):
     """Return the repeat/field-list rows to flatten, or ``None`` to keep as-is.
 
--- a/f/connectors/kobotoolbox/kobotoolbox_responses.py
+++ b/f/connectors/kobotoolbox/kobotoolbox_responses.py
@@ -395,6 +395,9 @@ def flatten_kobotoolbox_submission(submission):
     slash-separated keys. Field-list groups may arrive as a lone ``dict`` with
     slash keys. Both are expanded to ``{group}/{index}/{leaf_key}`` (1-based index).
 
+    Repeat groups can be nested (a repeat inside a repeat), so flattening recurses
+    until no slash-keyed rows remain, e.g. ``{outer}/{i}/{inner}/{j}/{leaf_key}``.
+
     Parameters
     ----------
     submission : dict
@@ -410,11 +413,21 @@ def flatten_kobotoolbox_submission(submission):
         rows = _get_flattenable_kobo_rows(value)
         if rows is None:
             flat[key] = value
-            continue
-        for index, item in enumerate(rows, start=1):
-            for slash_key, field_value in item.items():
-                flat[f"{key}/{index}/{slash_key.rsplit('/', 1)[-1]}"] = field_value
+        else:
+            _flatten_kobo_rows(key, rows, flat)
     return flat
+
+
+def _flatten_kobo_rows(prefix, rows, flat):
+    """Expand ``rows`` under ``prefix`` into ``flat``, recursing into nested repeats."""
+    for index, row in enumerate(rows, start=1):
+        for slash_key, value in row.items():
+            key = f"{prefix}/{index}/{slash_key.rsplit('/', 1)[-1]}"
+            nested = _get_flattenable_kobo_rows(value)
+            if nested is None:
+                flat[key] = value
+            else:
+                _flatten_kobo_rows(key, nested, flat)
 
 
 def transform_kobotoolbox_form_data(form_data, form_name=None, form_languages=None):
--- a/f/connectors/kobotoolbox/kobotoolbox_responses.py
+++ b/f/connectors/kobotoolbox/kobotoolbox_responses.py
@@ -342,6 +342,71 @@ def download_form_responses_and_attachments(
     return form_submissions, form_name, form_languages
 
 
+# TODO -YAGNI: Extract to f/common_logic/submission_flatten.py if/when ODK repeat/matrix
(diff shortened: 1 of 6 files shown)


f/connectors/kobotoolbox/kobotoolbox_responses.py
```diff
@@ -342,6 +342,70 @@ def download_form_responses_and_attachments(
     return form_submissions, form_name, form_languages
 
 
+# TODO - YAGNI: Extract to f/common_logic/submission_flatten.py if/when ODK
+# flattening is required; not worth a common module until then.
+def _slash_keyed_rows(value):
```
conservationtimothy: can we maybe rename the function to: `_get_kobo_nested_rows` or `_get_flattenable_kobo_value` and or change the comments, as it took me a while to understand that, in plain English:

```text
Given a Kobo field:

If it looks like:
    [
      {"some/path": value},
      {"some/path": value}
    ]

or:

    {
      "some/path": value
    }

then return it so we can flatten it.

Otherwise:
    leave it alone.
```

Obviously maybe not as simple as that but the comments were hard for me to parse too!

f/connectors/kobotoolbox/kobotoolbox_responses.py
```diff
@@ -342,6 +342,70 @@ def download_form_responses_and_attachments(
...
+        and value
+        and all(isinstance(k, str) and "/" in k for k in value)
+    ):
+        return [value]
+
+    # Scalars, empty containers, geolocation, attachments, and anything else.
+    return None
+
+
+def flatten_kobotoolbox_submission(submission):
```
conservationtimothy, line 391: Is it possible for Kobo submissions to contain deeply nested repeat groups (e.g. a repeat inside another repeat)? The current flattening logic only handles one level of slash-keyed rows, so I was wondering whether nested structures are something we need to account for, or whether Kobo's output shape guarantees this isn't needed.

Note: I saw the "if/when ODK flattening is required" TODO, so if this is more relevant to ODK rather than Kobo, feel free to ignore.

Not a blocker since I saw this working in practice, just curious about the expected payload shape.

f/connectors/kobotoolbox/kobotoolbox_responses.py
rudokemper: Thanks. I will push up an improvement.

f/connectors/kobotoolbox/kobotoolbox_responses.py
rudokemper: Great catch. Thanks for pointing this out. I tried it out with a real Kobo form and discovered that yes, you can have repeats inside of repeats, and we needed to implement recursive flattening to handle this. https://github.com/ConservationMetrics/gc-scripts-hub/pull/256/commits/980e25f5cda70db1df3c8a43ce6a5e98f0b610e6

conservationtimothy: Thanks for this

conservationtimothy (approved):
Approved but don't know if @IamJeffG might wanna give this the once over.

PR #257, opened 2026-07-02
Revise contributing for large PRs

IamJeffG:
## Goal

Revise `CONTRIBUTING.md` guidelines to acknowledge necessarily large PRs. 

## What I changed and why

It acknowledges that certain refactors are legitimately large.
- The solution is the guidelines now focus as much on cohesion of PRs as size.
- It provides a pattern of using a single commit-indexed PR to handle that (in `CONTRIBUTING_EXAMPLES.md`).

It also ramps up nudging toward pre-review consultation and communication with your reviewer.

This guidance comes from a series of conversations within our team, when we were doing a number of large refactors.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -51,9 +51,11 @@ As code author:
 - When you anticipate a PR will necessarily be large, loop in your reviewer early: they may have ideas that reduce scope, or they may want to agree in advance on a review approach.
 - The first reviewer of the code or documentation that you submit should be YOU!  (More at https://blog.beanbaginc.com/2014/12/01/practicing-effective-self-review/)
 - Ultimately the responsibility for bug-free code is on the code author, not the reviewer.
+- Orient your reviewer. Use direct outreach or the PR description to convey scope, flag what's
+  boilerplate or auto-generated, and highlight where you want focused attention.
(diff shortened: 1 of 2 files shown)


conservationtimothy:
> I wonder if now is the time to start checking-in AGENTS.md, CLAUDE.md to our repos. I've been including an edited version of https://github.com/multica-ai/andrej-karpathy-skills/blob/main/CLAUDE.md and seeing great success.
> 
> If we are in favor of this, I have a script that can add the same file to all our repos...

Would it be worthwhile to have a version which includes the best bits from all our individual `AGENTS.md` or equivalent files? So we can all share it privately or publicly and think about creating a hybrid.

rudokemper (approved):

conservationtimothy (approved):

JumaKahiga (approved):
I support having a standard `AGENTS.md` that can be customized for different repos where necessary. Beyond keeping the "no extra code beyond what is absolutely necessary" principle, I'd also use it to capture our general architecture conventions, coding patterns, and project-specific expectations.

Karpathy's template is a good place to start, but I have also discovered that brevity such as what the template goes for is not always the best choice with co-pilot tools :) . We don't need to get it perfect on day one. We can check in an initial draft and then evolve it through the normal PR review process as we discover what actually works for us.

PR #260, opened 2026-07-08
Use dynamic class to hide convert lat/lon to GeoJSON for non-CSV uploads

rudokemper:
## Goal

I learned about [dynamic styling](https://www.windmill.dev/docs/apps/app_configuration_settings/app_styling#dynamic-styling) in Windmill apps recently and figured we could take advantage of this feature to selectively hide the "Convert lat/lon fields to GeoJSON?" components if it's not a CSV upload.

## Screenshots

If CSV:

image

If non-CSV:

image

## What I changed and why

- Add a dynamic styling expression `state.outputFormat !== "csv" ? "hidden" : ""`.
- The rest of the changes are just noise

Trying it in runtime (as per screenshots)

## What I'm not doing here

You might be wondering if we can vertically center the "Is this dataset from a known source or tool?" section, if the conversion stuff is hidden. I think that would be tricky, with responsive design considerations. I think the negative space is an acceptable tradeoff for lessened mental burden of not needing to think about this section if it's not relevant.

## LLM use disclosure

None

### DIFF
--- a/f/apps/gc_dataset_importer.app/app.yaml
+++ b/f/apps/gc_dataset_importer.app/app.yaml
@@ -30,6 +30,10 @@ value:
       container:
         class: ''
         style: ''
+    selectcomponent:
+      input:
+        class: ''
+        style: ''
     steppercomponent:
       container:
         class: ''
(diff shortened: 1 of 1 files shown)


nicopace (approved):
There it is.
For some reason I couldn't find the button the last time (mental melting/github trickery).

PR #267, opened 2026-07-28
Remove "if file exists then don't download" check for ArcGIS anonymous download script

### LINKED ISSUE #263: ArcGIS download feature layer script is not overwriting files that already exist on disk
One of our partners uses this script to pull data from a public data source that gets updated regularly.
But if the files are there, it doesn't update them.
https://github.com/ConservationMetrics/gc-scripts-hub/blob/f9c21340cbe01fc36809fe28675f5dfe82d4f685/f/connectors/arcgis/arcgis_download_feature_layer_anonymously.py#L348

It should update the files, or have a setting to configure this behaviour at least.

rudokemper:
## Goal

Closes https://github.com/ConservationMetrics/gc-scripts-hub/issues/263.

@nicopace, I didn't realize it was the script that you yourself authored that needed a fix! 

If you want to make the fix directly next time and submit a PR, that is most welcome. (Not just limited to this script, I'm happy to work with you to get your fixes merged in, in accordance to the standards set in [`README.md#development`](https://github.com/ConservationMetrics/gc-scripts-hub/blob/main/README.md#development)

## What I changed and why

- Removed "if filename exists then don't skip download check" in anonymous arcgis download script.
- Added a test for this.

Tested in runtime.

## What I'm not doing here

- The `arcgis_feature_layer.py` script already matched this behavior. So it was only the `arcgis_download_feature_layer_anonymously.py` script that needed an adjustment.

## LLM use disclosure

None

### DIFF
--- a/f/connectors/arcgis/arcgis_download_feature_layer_anonymously.py
+++ b/f/connectors/arcgis/arcgis_download_feature_layer_anonymously.py
@@ -392,10 +392,6 @@ def fetch_layer_data(
     filename = storage_path / f"{layer_name}.{output_format}"
     relative_output = Path(storage_path.name) / f"{layer_name}.{output_format}"
 
-    if filename.exists():
-        logger.info("File %s already exists. Skipping download.", filename)
-        return relative_output
-
     base_feature_url = f"https://{subdomain}.arcgis.com/{service_id}/arcgis/rest/services/{feature_id}/FeatureServer"
 
     records = fetch_features(session, base_feature_url, layer_index)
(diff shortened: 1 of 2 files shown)


nicopace (changes requested):
The implementation looks ok.

would it work if the fetched file responds 404? Or empty file?
What do you think if we consider this usecase? Not a dealbreaker, but I imagine that it could happen, and dropping the current snapshot might mean we might not recover the data again

nicopace (approved):

PR #273, opened 2026-08-07
Add a Mapbox script to download, and reconstruct, vector tiles

### LINKED ISSUE #272: Add a Mapbox vector download and stitching script
Base it on https://github.com/ConservationMetrics/gc-programs/issues/190

The code in question is this one: https://github.com/ConservationMetrics/gc-programs/blob/main/scripts/mapbox/download_mapbox_data.py

And there is a companion README next to it.

rudokemper:
## Goal

Closes https://github.com/ConservationMetrics/gc-scripts-hub/issues/272.

## Screenshots

image

## What I changed and why

- Basically, I took @nicopace's clever and fully functional [`download_mapbox_data.py`](https://github.com/ConservationMetrics/gc-programs/blob/main/scripts/mapbox/download_mapbox_data.py) script in `gc-programs` and turned it into a production-grade script for `gc-scripts-hub` with Windmill parameterization and tests.
- Some changes I made to the script:
  - Used `mercantile` to convert lat/lon to tile coordinates instead of custom `deg2num` fn
  - In addition to `204` status code, I also discovered that the Mapbox API can return `404` for when there is no coverage outside the tileset extent. So for this we cache an empty marker so the cell is never re-requested.
  - Got rid of `max_zoom` arg, handle validation in `zoom` instead.
  - Instead of `tileset` being an Array, I think it's better to run this script tile-by-tile. As the `reconstruct_column` might differ from source to source. So I think the mental model is better for this to be 1 run per tileset.
  - Implemented other repo-wide standards like use of `logging`, `pathlib`, and our `save_data-to_file` library to handle saving as GeoJSON.
- Lastly, I discovered that `minimum` and `maximum` conditions don't work for Integer parameters (i.e. what was being used for zoom level). So instead I set those fields -- for both mapbox scripts -- to String, and used a `validation` instead to validate a proper zoom level and ceiling.

- Trying it in Windmill runtime.
- See https://github.com/ConservationMetrics/gc-programs/issues/190 for partner-based motivation on the original script, and why we've added it here.

## LLM use disclosure

I used Cursor Grok 4.5 to Windmill-ize the existing script and apply `gc-scripts-hub` coding patterns, and generate tests. Then I made various changes to improve the code, like implementing `mercantile` and streamlining the script params.

### DIFF
--- a/f/connectors/mapbox/README.md
+++ b/f/connectors/mapbox/README.md
@@ -1,3 +1,7 @@
+# Mapbox connectors
+
+Scripts for working with Mapbox tilesets and vector tile data.
+
 # `mapbox_create_or_update_tileset`: Create or Update a Mapbox Tileset
 
 This script uses the Mapbox Tiling Service (MTS) to create **or** update a Mapbox tileset from a GeoJSON file by:
@@ -29,7 +33,7 @@ This script is intended for workflows where a GeoJSON dataset is updated regular
 
 ## Mapbox Secret Access Token
 
-For the scripts to work, you need to provide a Mapbox secret access token with scope to work with tilesets. You can create a new secret access token in Mapbox Studio by:
+For this script to work, you need to provide a Mapbox secret access token with scope to work with tilesets. You can create a new secret access token in Mapbox Studio by:
 
 1. Navigating to **Admin >  Tokens**
 2. Clicking **+ Create a token**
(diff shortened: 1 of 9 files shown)


nicopace (approved):
looks good!

conservationtimothy (approved):

PR #275, opened 2026-08-18
Add a Sensing Clues connector

rudokemper:
## Goal

To add a connector script for [Sensing Clues](https://sensingclues.org/).

## Screenshots

image

_Sensing Clues data pulled in by this script, visualized in GC Explorer._

## What I changed and why

I added `f/connectors/sensingclues/sensingclues_observations.py` plus the usual scaffolding (Windmill YAML, fixtures from the public demo projects, tests, README).

Things of note (and to emphasize for review as the rest follows similar patterns):

- We use the `sensingcluespy` client for login, cookies, and `make_query()`, same idea as how we use pyODK in the ODK Connector. We deliberately do **not** use `get_observations()`, as it explodes one row per concept and drops attributes.
- Form `attributes` flatten into GeoJSON properties; ontology concepts collapse into `conceptLabels` / `conceptIds` lists. Geometry comes from `Observation.where`.
- Username / password are plain params (no resource type). 

I tested it against the public demo account (`demo` / `demo`) for both demo groups. Pagination on the 2414-record group came back with 2414 unique IDs. `tox -e sensingclues` is green.

## What I'm not doing here

What I mentioned as additional work in the README.

## LLM use disclosure

Plan by Cursor Opus 4.9 after a prompt with explicit instructions, API references, and reviewing README and repo content for reference. Implementation by Cursor Grok 4.5. Then review and various edits by me.


f/connectors/sensingclues/sensingclues_observations.script.lock
```diff
@@ -0,0 +1,25 @@
+# workspace-dependencies-mode: manual
+# py: 3.11
+# Must stay Python 3.11. `wmill script generate-metadata` targets 3.12 and pins
+# numpy==2.5.2, which will not install on CircleCI/tox (3.11).
```
IamJeffG: is this your code comment? I assume it'll get overwritten by some future run of `windmill generate-metadata`

Then again, I don't know what workspace-dependencies-mode is...

IamJeffG: is this your code comment? I assume it'll get overwritten by some future run of windmill generate-metadata

Then again, I don't know what workspace-dependencies-mode is...

rudokemper: You're right. I've moved that comment to the actual Python code and specified the version there (per https://www.windmill.dev/docs/getting_started/scripts_quickstart/python#select-python-version), which will ensure `wmill script generate-metadata` will use py311.

I am not sure why Windmill injected `workspace-dependencies-mode` and [the relevant doc page](https://www.windmill.dev/docs/core_concepts/workspace_dependencies) doesn't clarify. But with https://github.com/ConservationMetrics/gc-scripts-hub/pull/275/commits/d1618a53184b8df45139a23a2f52d5ce1babb263 which pins py311 in the code, and then running `wmill script generate-metadata`, it didn't add it again. So I removed it.

IamJeffG:
Requesting @Atmosfearful to be the primary reviewer on this.

f/connectors/sensingclues/README.md
```diff
@@ -0,0 +1,34 @@
+# Sensing Clues
+
+[**Sensing Clues**](https://sensingclues.org/) is a wildlife monitoring platform. Rangers and community members record observations in the [Cluey](https://www.sensingclues.org/data-collection) Data Collection app; these are then synchronized with Sensing Clues, which stores those records as observations tagged with an ontology of species, activities, and other concepts. Users can access these observations via the [Sensing Clues Central](https://central.sensingclues.org/) web application, where they can also upload CSVs.
+
+## `sensingclues_observations.py`
+
+Fetches observations for one "Group" (Sensing Clues' equivalent of a project) via the [sensingcluespy](https://sensingcluespy.readthedocs.io/en/latest/) client. Saves raw JSON and GeoJSON to the datalake and writes features to PostgreSQL.
+
+Each observation becomes one GeoJSON Feature. Ontology concepts collapse into `conceptLabels` / `conceptIds` lists. Form fields in `attributes` flatten into properties (core fields such as `fileName` and `tags` win on collision). Geometry comes from `Observation.where`.
```
Atmosfearful: It's not possible to collect non-geodata with Cluey? Like generic questionnaires / forms?

f/connectors/sensingclues/sensingclues_observations.py
```diff
@@ -0,0 +1,287 @@
...
+)
+
+
+logging.basicConfig(level=logging.INFO)
+logger = logging.getLogger(__name__)
+
+
+def main(
+    username: str,
+    password: str,
```
Atmosfearful: Are you aware of any ambient logging happening in windmill or elsewhere (like script invocation logs?) that might potentially lead to this PW being persisted in plaintext?

I have no reason to believe either way, but it's worth asking.

f/connectors/sensingclues/tests/conftest.py
```diff
@@ -0,0 +1,149 @@
...
+    return SensingCluesServer(
+        username="bad",
+        password="wrong",
+        group_identifier=server_responses.CLUEY_IDENTIFIER,
+    )
+
+
+@pytest.fixture
+def pg_database():
+    db = testing.postgresql.Postgresql(port=7654)
```
Atmosfearful: Please ask an LLM to merge in changes from `main` and use the new cached factory without the fixed port, so we can keep leveraging the parallel tox runners

Atmosfearful (changes requested):
## My Human findings;
1 question, and 1 change request below in comments.

## AI findings;

### Attribute-name collisions corrupt column semantics
`f/connectors/sensingclues/sensingclues_observations.py:260-267`  
`f/connectors/geojson/geojson_to_postgres.py:25-31`

_[Human note: I believe this touches shared code, so fix may be out of scope for this specific PR?]_


SensingClues form attributes become individual database columns, but form creators can use labels that are different to users while becoming identical after conversion to PostgreSQL-compatible names. The same source field must always map to the same database column for the table to remain trustworthy.

The generic writer normalizes every row's keys independently. Distinct source fields that normalize to the same SQL identifier can consequently be assigned different SQL columns in different records, silently mixing unrelated values.

The root cause is in shared code: `StructuredDBWriter.handle_output()` in `f/common_logic/db_operations.py:673-685` does not retain the updated key mapping returned by `sanitize_sql_message()` between rows.
So it is not unique to SensingClues.
SensingClues specifically exposes it because:
- It flattens arbitrary API `attributes[].key` values into feature properties.
- It persists through `geojson_to_postgres.py`.
- That shared GeoJSON writer explicitly sets use_mapping_table=False.
Other potentially affected paths include:
- Any connector using `geojson_to_postgres`, because it shares the same `use_mapping_table=False` writer configuration.
- `auditor2` and `timelapse`, which also explicitly disable mapping tables.
- `csv_to_postgres` when called with `use_mapping_table=False`.
Connectors using `use_mapping_table=True`, such as ODK, EpiCollect, and KoboToolbox, are intended to avoid cross-run mapping drift. However, the shared writer still needs to retain newly created mappings while processing a batch; simply opting into the mapping table does not by itself prove this exact same-batch collision path is fixed.


### Empty imports retain stale published data
`f/connectors/sensingclues/sensingclues_observations.py:178-205`

*[Human note: I believe another way to rephrase this issue here is, "user deletes data upstream, leading to an empty import; previously imported data is not cleaned up in this scenario"]*

Each run appears to download the current observations for the requested group and write those results to the named outputs. A successful query that returns no observations should not leave consumers believing the previous query results are still current.

An empty response skips raw-file, GeoJSON, and database writes. If an earlier run populated the same table, consumers continue seeing old observations despite the successful current query having no results.

**Proposed solution:** Decide and document whether this connector is snapshot-based or incremental. For snapshots, write an empty raw archive and empty FeatureCollection, then atomically clear or replace the target database table. If incremental behavior is intended, preserve current data deliberately but make that explicit and ensure date-filtered runs cannot be mistaken for snapshots.

f/connectors/sensingclues/sensingclues_observations.py
rudokemper: Hm, well, it sure is in the Run history (retained for 2 weeks on non-enterprise Windmill self-hosted deployments).

image

I will resolve this by adding
```json
      password: true
``` 

to the password param.

rudokemper:
Thanks, @Atmosfearful! I addressed your two code review comments, and made an issue for the 2nd AI-generated one.

RE: `Attribute-name collisions corrupt column semantics`. I think I agree this should be investigated. Could you open an issue to containerize it? I do think it's scope drift for the work here as it requires a repo-wide audit and modification.

f/connectors/sensingclues/sensingclues_observations.script.yaml
```diff
@@ -61,6 +61,7 @@ schema:
       description: Sensing Clues Focus password.
       default: null
       originalType: string
+      password: true
```
Atmosfearful: good to know!

Edit: I can't find any docs on how this works but I also didn't look very hard. Assuming it's a windmill thing? It's the perfect example of a rule that belongs in a repo-specific AGENTS.md. For now I'll put it in my workspace one which is still evolving.

Atmosfearful (approved):

f/connectors/sensingclues/sensingclues_observations.script.yaml
rudokemper: Yep that's it. And I agree! That would be super useful.

PR #276, opened 2026-08-19
Fix Dataset Importer Excel values, save gating, and column-name sanitization

rudokemper:
## Goal

Make Dataset Importer round-trip spreadsheet data more faithfully and actually allow updates to existing tables.

## What I changed and why

- Excel (and similar tabular) values that are whole numbers are no longer stringified with a trailing `.0`, and and empty cells become "" instead of "nan".
- Dataset Importer **Save & Finish** is enabled when there are new rows, updated rows, or new columns, so re-uploads of an existing dataset are no longer blocked.
- Dataset Importer column names now keep word spaces as underscores and reverse `/`-separated group/question paths the same way Kobo does (`Basic information/Are you married?` → `Are_you_married__Basic_information`).

Working with cleaned up Kobo data from a partner and noticing these discrepancies and issues.

## LLM use disclosure

Very informed prompting to Cursor Grok 4.5 to make these changes, and add tests.


Atmosfearful:
Oops just noticed this got lost in the shuffle, I will start review now!

Atmosfearful (approved):

PR #278, opened 2026-08-26
Remove tox-docker and add gcp_storage_emulator directly to pytest runs

### LINKED ISSUE #277: Remove Docker dependency for test runs
## Background
For the `tox` test runs to pass, the project requires `docker` to be installed on the runner.

## Problems
To run tests inside of a containerized workspace (devpod / devcontainer etc.), we would need to A) install docker-in-docker, or B) expose the host machines `docker.sock`

Option B breaks the security boundary and would defeat the purpose of using a containerized workspace (which I would very much like to keep for agent isolation + to limit splash from supply chain attacks), so that's off the table IMO.
Option A is doable since this is just for test runs, but adds bloat and memory overhead so is not ideal.

Loosely related, the docker image's version today isn't pinned which could introduce test flakiness or drift.

## Solution
As I understand it, the docker dependency comes from the use of `tox-docker` which we use to run `gcp-storage-emulator`.

So option C) is to just install that emulator directly and remove the `tox-docker` dependency and related config.
Test coverage should not change, no runtime code would be touched, might even see a tiny speedup to CI runs.

Other research notes:
```
gcp-storage-emulator introduces no new host-level dependencies.
It requires only:
Python 3.10+
The Python package google-crc32c, which the alerts environment already installs
A loopback TCP port inside the devcontainer
```

Atmosfearful:
## Goal
Make repo more compatible with containerized workspaces (devcontainer.json).

More details in the issue - #277 

## Screenshots
No visual changes

## What I changed and why
Removed tox-docker from configs and readme.
Updated conftest.py to directly invoke `gcp_storage_emulator` in memory, rather than going over the docker network.

Same tests pass without refactor

## What I'm not doing here
No runtime changes, No change in test coverage nor major test refactor

## LLM use disclosure
GPT 5.6 Sol via OpenCode

### DIFF
--- a/f/connectors/alerts/tests/conftest.py
+++ b/f/connectors/alerts/tests/conftest.py
@@ -1,7 +1,6 @@
-import os
-
 import pytest
 import testing.postgresql
+from gcp_storage_emulator.server import create_server
 from google.auth.credentials import AnonymousCredentials
 from google.cloud import storage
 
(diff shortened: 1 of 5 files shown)


f/connectors/alerts/tests/conftest.py
```diff
@@ -18,23 +17,21 @@ def pg_database():
...
-    # Create a storage client that connects to the emulator
-    storage_client = storage.Client(
-        project="test-project",
-        credentials=AnonymousCredentials(),
-        client_options={"api_endpoint": emulator_endpoint},
-    )
-
-    yield storage_client
-
-    # Clean up: Delete all buckets
```
IamJeffG: I believe this might've been here because successive runs could have kept data.  I assume in a docker volume.

Where does the python library store its buckets and is there a chance if it not getting cleared run-over-run?

IamJeffG (commented):
LGTM but I'd like @rudokemper to approve.

f/connectors/alerts/tests/conftest.py
Atmosfearful: My understanding is that the new setup already clears itself between test runs as we invoke `create_server` right in the function scope now (with `in_memory=True` flag on L23), and close it in the teardown after `yield`

rudokemper (approved):
Great! It didn't occur to me that we only used tox-docker for the alerts tests. (Probably because alerts is one of the first two connectors we worked on long ago, so back then it was used in 50% of the repo 🙂)

rudokemper:
@Atmosfearful please ignore the Github assigning / automation noise for now - I will walk you through our PM process when we meet.

PR #279, opened 2026-08-27
Test Speedup #2 - Run tox environments with two workers

Atmosfearful:
Part 2 of 3 the other two PRs are #281 and #280

Benchmark:
[After this PR](https://app.circleci.com/pipelines/github/ConservationMetrics/gc-scripts-hub/853/workflows/cc09e7c9-605b-4f67-8b02-a0f9a3ca228e/jobs/852) - 3m 9s (shaves ~2m 29s off of each run)

Coincidentally the same amount of time reduced by #280 

**Edit 2026.08.31**

Snuck another speedup in here, I noticed that we weren't caching dependencies so I rewrote config.yml to leverage circleci images.
Slowest observed run, before all 3 PRs: ~9 mins.
Fastest observed run, the most recent commit: 2m 21s

The next obvious speedups that I am aware of would be either to the python tests themselves, or to avoid the `postgres` install. The latter could be done with a sidecar postgres image (better, but that changes how we develop locally, so i avoided it).

It's worth investigating at some point. That postgres install can be flaky, I noticed one build in our logs taking more than 30 minutes just to install postgres. I tried to partially mitigate it by targeting postgres-14 (the only version available for Ubuntu Jammy) and disabling 'recommends'.

## Goal
Before this PR we couldn't use `tox -p 2` to run two test workers in parallel.
The reason, citing GPT-5.6 Sol
> With tox -p 2, two environments may each try to start PostgreSQL simultaneously, and one fails with “address already in use.”
> This branch changes it to:
> testing.postgresql.Postgresql()
> When no port is supplied, testing.postgresql chooses an available port automatically. Each test’s DSN then contains its own port, so concurrent PostgreSQL instances do not collide.

This will create merge conflicts with with changes in #281 , but I separated them to make sure we have a benchmark comparison.

## Screenshots
No visual changes

## What I changed and why
Remove port hardcoding from `testing.postgresql.Postgresql()`

Passing tests

## What I'm not doing here
Touching runtime code

## LLM use disclosure
GPT-5.6 Sol and Terra

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -47,7 +47,7 @@ jobs:
             pip install tox~=4.18 tox-docker~=5.0
       - run:
           name: Run tests
-          command: tox -- -v
+          command: tox -p 2 -- -v
           environment:
             PIP_UPLOADED_PRIOR_TO: P7D
             # Skip tox-docker's NetworkSettings.Gateway lookup (broken on newer
--- a/f/common_logic/tests/conftest.py
+++ b/f/common_logic/tests/conftest.py
@@ -11,7 +11,7 @@ from f.common_logic.db_operations import conninfo
 @pytest.fixture(scope="function")
 def _postgres_instance():
     """Spin up a temporary PostgreSQL instance for tests."""
-    db = testing.postgresql.Postgresql(port=7654)
+    db = testing.postgresql.Postgresql()
     try:
         yield db
     finally:
--- a/f/connectors/alerts/tests/conftest.py
+++ b/f/connectors/alerts/tests/conftest.py
@@ -9,11 +9,11 @@ from google.cloud import storage
 @pytest.fixture
 def pg_database():
     """A dsn that may be used to connect to a live (local for test) postgresql server"""
-    db = testing.postgresql.Postgresql(port=7654)
+    db = testing.postgresql.Postgresql()
     dsn = db.dsn()
     dsn["dbname"] = dsn.pop("database")
     yield dsn
-    db.stop
+    db.stop()
 
 
 @pytest.fixture
--- a/f/connectors/arcgis/tests/conftest.py
+++ b/f/connectors/arcgis/tests/conftest.py
@@ -154,8 +154,8 @@ def arcgis_anonymous_server(mocked_responses):
 @pytest.fixture
 def pg_database():
     """A dsn that may be used to connect to a live (local for test) postgresql server"""
-    db = testing.postgresql.Postgresql(port=7654)
+    db = testing.postgresql.Postgresql()
     dsn = db.dsn()
     dsn["dbname"] = dsn.pop("database")
     yield dsn
-    db.stop
+    db.stop()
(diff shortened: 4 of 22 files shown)


IamJeffG (approved):

rudokemper (approved):
Nice improvement!

Atmosfearful:
Final update, updated PR description with notes;


> Edit:
> 
> Snuck another speedup in here, I noticed that we weren't caching dependencies so I rewrote config.yml to leverage circleci images.
> Slowest observed run, before all 3 PRs: ~9 mins.
> Fastest observed run, the most recent commit: 2m 21s
> 
> The next obvious speedups that I am aware of would be either to the python tests themselves, or to avoid the `postgres` install. The latter could be done with a sidecar postgres image (better, but that changes how we develop locally, so i avoided it).
> 
> It's worth investigating at some point. That postgres install can be flaky, I noticed one build in our logs taking more than 7 minutes just to install postgres.
>

rudokemper:
I didn't know about parallel runs in tox before this PR, and so I tried it locally with current `main`:

- `$ tox` took 1:16 (a _massive_ improvement from before, great work @Atmosfearful 
- `$ tox -p 2` took 0:44
- `$ tox -p 5` took 0:33 (my laptop fan started to groan a bit for this one!)

Two parallel workers at least seems like a good win worth documenting as a `> [!TIP]` in the README, I think? Like a quick blurb "If you want to speed up running tests locally, you can use the parallel run `-p` flag to increase the number of workers"

.circleci/config.yml
```diff
@@ -17,34 +10,23 @@ workflows:
...
-          # Additionally, any tests for scripts that use the Azure Storage Blob SDK
-          # fail on CircleCI with Python 3.13, even though the scripts run fine in
-          # Windmill runtime with 3.13.
-          # TODO: Figure out why the Azure Storage Blob SDK fails on CircleCI with 3.13
-          name: Install and set Python 3.11
-          command: |
-            pyenv install 3.11.10
-            pyenv global 3.11.10
+            sudo apt-get update
+            sudo apt-get install --yes --no-install-recommends postgresql-14
```
IamJeffG: Why such an old version?

Atmosfearful: I stuck with the image for the same python version we were targeting before, but that image is using Ubuntu Jammy which only had pg 14. I figured since postgres is pretty good at backwards compat it wouldn't hurt, and would be lower risk than trying to bump python (which we should do eventually). Not sure how good the python world is about compatability and wanted to avoid scope creep

rudokemper: > trying to bump python (which we should do eventually)

For the record, we pinned (and actually downgraded) Python version to 3.11 in https://github.com/ConservationMetrics/gc-scripts-hub/pull/124 because of incompatibility issues between 3.13 and Azure Blob Storage SDK dependencies. Those might be resolved by now. I think it would be good to create an issue to track doing this work.

Atmosfearful: https://github.com/ConservationMetrics/gc-scripts-hub/issues/282

PR #280, opened 2026-08-27
Test speedup #3 - Remove delay after PostgreSQL mapping writes

Atmosfearful:
Part 3 of 3 the other two PRs are #281 and #279 

Note, this one touches runtime code that I am not familiar with yet.

My understanding is based on this conversation (use the nav on the left to see my follow-up questions)
https://opncd.ai/share/n5HdvkkQ

So the super short summary is;
- was ported from older dagster code
- later on, we changed the db connection to to 'autocommit=True'
- this sleep was probably necessary only in the old code, where statements were not committed synchronously.

Benchmark:
[After this PR](https://app.circleci.com/pipelines/github/ConservationMetrics/gc-scripts-hub/852/workflows/106bccf6-b39d-4436-abaa-c1607481e92b/jobs/851) - 3m 9s (shaves ~2m 29s off of each run)

## Goal

## LLM use disclosure
GPT-5.6 Sol and Terra

### DIFF
--- a/f/common_logic/db_operations.py
+++ b/f/common_logic/db_operations.py
@@ -7,7 +7,6 @@ This module provides functions and classes for interacting with PostgreSQL datab
 
 import json
 import logging
-import time
 
 from psycopg import Error, connect, errors, sql
 
@@ -719,7 +718,6 @@ class StructuredDBWriter:
                 self._create_missing_mappings(
                     pgconn, columns_table_name, missing_mappings
                 )
-                time.sleep(10)
 
             inserted_count = 0
             updated_count = 0


rudokemper:
Just to validate the LLM's analysis that this was ported over, here is [the same delay code](https://github.com/ConservationMetrics/frizzle-dagster/blob/cd89c04bf58c819466a074fd8ebb3f13df0bfa73/dagster/core/core/postgres_io.py#L576-L578) in our actual legacy Dagster implementation. It was introduced in [#27](https://github.com/ConservationMetrics/frizzle-dagster/pull/27), a PR that had a **lot** going on, so not worth trying to comb through to find rationale. (I actually already tried.)

I don't remember the context for why we had a manual delay after Postgres mapping writes -- @IamJeffG might -- but it does sound like an artifact of us wrestling with Dagster, and I agree that it seems superfluous in this repo. But I'm going to defer to Jeff on approving this as he was more involved in the original Postgres write code than I.

IamJeffG (approved):
I know no reason this `sleep` is supposed to be there.

PR #281, opened 2026-08-27
Test speedup #1 - Cache initialized PostgreSQL databases in tests

Atmosfearful:
Part 1 of 3 - the other two PRs are #279 and  #280 

Benchmark:
[After this PR](https://app.circleci.com/pipelines/github/ConservationMetrics/gc-scripts-hub/851/workflows/35a43c17-bf70-436a-8141-f6413be83dbe/jobs/850) - 4m 49s (shaves ~49s off of each run)

## Goal

Before this PR, tests were invoking postgresql.Postgresql() directly.
Apparently, this is invoking postgres `initdb` under the hood, which is expensive because it creates an entire PostgreSQL cluster, when all we need is an empty db for the test block

After this PR, each test now relies on PostgresqlFactory, which lets us init once as a template, and reuse it to copy and start a new postgres server/process for each test.

The parent fixture that invokes the factory is responsible for cleanup (db.stop()). Note: Many tests are **not properly invoking stop()** (they were referencing it as `db.stop` instead of invoking it). These will get fixed in #279 

## Screenshots
No visual changes

## What I changed and why
See above

Passing tests

## What I'm not doing here
No runtime changes

## LLM use disclosure
GPT-5.6 Sol and Terra

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -1,0 +1,14 @@
+import pytest
+
+
+@pytest.fixture(scope="session")
+def postgresql_factory():
+    import testing.postgresql
+
+    factory = testing.postgresql.PostgresqlFactory(
+        cache_initialized_db=True, port=7654
+    )
+    try:
+        yield factory
+    finally:
+        factory.clear_cache()
--- a/f/common_logic/tests/conftest.py
+++ b/f/common_logic/tests/conftest.py
@@ -3,15 +3,14 @@ import zipfile
 from pathlib import Path
 
 import pytest
-import testing.postgresql
 
 from f.common_logic.db_operations import conninfo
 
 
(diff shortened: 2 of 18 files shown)


IamJeffG (approved):

conftest.py
```diff
@@ -0,0 +1,14 @@
+import pytest
+
+
+@pytest.fixture(scope="session")
+def postgresql_factory():
+    import testing.postgresql
```
rudokemper: Before merging, maybe add a docstring describing what this is and how it works with each specific fixture? (You already go into some detail about this in your PR body, it would be helpful to have it here for reference.)

rudokemper (approved):

PR #283, opened 2026-09-02
Add tip about tox parallel workers

rudokemper:
## Goal

Self-explanatory, I think! But pulling out an insight from @Atmosfearful recent PR on speeding up `tox` tests in CI.

## LLM use disclosure

None


Atmosfearful (approved):
Thanks for adding this,
yes on my desktop -p 8 got it down to like 30 seconds 🔥 

I added a similar line to a Validation `SKILL.md` file in my devcontainer workspace as well that I will share at some point (when i'm confident it's stable)

PR #291, opened 2026-09-14
Fetch iNaturalist observations via API v2 with an explicit field spec for research grade occurrences

### LINKED ISSUE #284: Audit iNaturalist connector fields for Scientific Grade occurrences
## Feature Request

Let's audit the iNaturalist connector code to ensure we download minimally essential fields for Scientific Grade occurrences. I am writing this quickly and can expound more, but I believe that we write the occurrence fields that are immediately given but dump a giant JSON with extra JOIN fields on disk that we might want to include in the record for each.

rudokemper:
## Goal

Closes #284.

Make the iNaturalist connector return a curated, inspectable observation archive: the fields we actually store, including Research Grade determinants, public geoprivacy, GBIF links, and sound attachments.

> [!NOTE]
>
> 806 / 1,119 new lines are from the fixture. 

## Screenshots

image

image

_Same dataset, but left pulled by script on `main`, right pulled by script from this PR._

## What I changed and why

Observation pulls now go through **API v2** with `_OBSERVATION_FIELDS` as the single source of truth for both the request and the table columns. v2 silently ignores unknown field names (HTTP 200, no error), so an implicit “whatever v1 returned” contract was unsafe.

Project metadata still uses **API v1** (`/projects/{slug}`). v2’s projects endpoint rejects slugs (`422 must be integer`).

The GeoJSON/Postgres row now includes:

- identity and provenance (`uuid`, `uri`, observer ORCID, `license_code`, `gbif_occurrence_id` from `outlinks`)
- Research Grade / quality fields (`quality_grade`, `captive`, taxon rank, identification agreement counts, `photo_count` / `sound_count`)
- public location quality (`obscured`, `geoprivacy`, `taxon_geoprivacy`, `positional_accuracy`, `public_positional_accuracy`) so fuzzed points are visible even without auth
- media: first photo unchanged (`photo_filename` / `photo_url`), plus `photo_filenames` / `sound_filenames`; sounds download as `{sound_id}.{ext}` next to photos

On-disk `{db_table_name}_observations.json` is this curated v2 archive. 

- `test_transform_keys_are_in_field_spec` asserts every transformed property is in `_OBSERVATION_FIELDS`. The project e2e checks a GBIF-linked research-grade row, an obscured record, a captive record, a sound-only record (no photo), and a `needs_id` row with no GBIF id.
- The fixtures actually reflect what is returned from a real iNaturalist project.
- I tried this in runtime, see screenshots.

## What I'm not doing here

No attempt to fetch every v2 field, only the spec. JSON on disk is no longer a raw dump.

## LLM use disclosure

Cursor Opus 5 drafted the plan. Cursor auto implemented the v2 field spec and encoder, the transform/media changes, the new tests, and README updates. I improved the fixture, reviewed and edited.


f/connectors/inaturalist/README.md
```diff
@@ -4,45 +4,61 @@
...
+- Pagination uses observation ID cursors (`id_above`) rather than page numbers, as recommended by iNaturalist for large result sets.
+- The script stays at or below ~60 requests per minute between paginated API calls, and pauses briefly between media downloads.
+- Photos are saved as `{photo_id}.{ext}` and sounds as `{sound_id}.{ext}` under `attachments/`.
+- Observations without visible coordinates are still stored with null geometry.
 
 ## Future work: supporting private or obscured coordinates
 
 Supporting private or obscured coordinates would require registering an iNaturalist application, completing OAuth2 to obtain an access token, exchanging it for a JWT via `/users/api_token`, and sending that JWT on API requests. JWTs expire after about 24 hours, so a long-lived integration would need refresh logic.
 
 ## 📚 Reference
```
Atmosfearful: Just to set a good example I would reference the inaturalist license and/or terms if they have it, and maybe a quick disclaimer, to assure that our use of the API is compliant. I only looked briefly but it seems that the only major restriction is on commercial AI training

https://www.inaturalist.org/pages/terms

f/connectors/inaturalist/inaturalist_pull.py
```diff
@@ -261,67 +340,96 @@ def _photo_url(observation: dict) -> str | None:
...
+            if url and filename:
+                yield _sized_photo_url(url, "original"), filename
+        for sound in observation.get("sounds") or []:
+            url = sound.get("file_url")
+            filename = _media_filename(sound, "file_url")
+            if url and filename:
+                yield url, filename
+
+
+def download_observation_media(
```
Atmosfearful: if it's a public function I would write tests for it, if it's private i would use `_` prefix convention... (actually I would write tests for even the simple private util functions too like `_iter_media` but it seems like that's a matter of opinion)

Atmosfearful (approved):

f/connectors/inaturalist/inaturalist_pull.py
rudokemper: Great feedback, thanks. Adding tests for both!

PR #292, opened 2026-09-14
Add bbox query parameter to iNaturalist script

### LINKED ISSUE #287: Modify iNaturalist script to also support use of bounding box to query API

rudokemper:
## Goal

Closes https://github.com/ConservationMetrics/gc-scripts-hub/issues/287.

## Screenshots

image

## What I changed and why

`slug` is optional. A new optional `bounding_box` string filters observations to a viewport.

- slug only → existing project/user query (unchanged)
- bbox only → `swlat` / `swlng` / `nelat` / `nelng` on the same v2 observations request; no project lookup
- both → both filters on one request
- neither, one corner, bad JSON, out-of-range, or a degenerate box → `ValueError`

Even though iNaturalist itself treats each coordinate as a separate query param, I opted to do a string field, to match what we do for e.g. GFW, and so we can point users (in README) to the Mapbox Location Helper for copying viewport bounds and make it easy to copy and paste once rather than 4 times.

- Expanded unit and e2e tests to cover the above cases
- Experimentation in runtime

## LLM use disclosure

Cursor implemented the bbox parameter, YAML changes, tests, and README based on my prompt about correct API usage and patterns to follow. I reviewed and edited.


f/connectors/inaturalist/inaturalist_pull.script.yaml
```diff
@@ -33,24 +36,35 @@ schema:
       default: null
       originalType: string
       pattern: "^.{1,54}$"
+    bounding_box:
+      type: string
+      description: >-
```
Atmosfearful: would be super helpful if you could link to a free tool or hint at an easy way to capture that bbox! Since we aren't rendering anything interactive

Atmosfearful (approved):
nice, good coverage, useful addition, and handles all the cases I can think of (invalid bbox)

PR #299, opened 2026-09-15
Resolve minor bugs and quality-of-life improvements for iNaturalist script

rudokemper:
## Goal

After running our iNaturalist script in production, I noticed a few things that required fixing:

- Windmill can timeout for huge iNaturalist jobs. So I made a note about that in the README.

- There was a minor bug where the fetch progress logs were decreasing the total number every time, e.g. 
    ```
    INFO:f.connectors.inaturalist.inaturalist_pull:[observations] Fetched 200 of 11438 observations
    INFO:f.connectors.inaturalist.inaturalist_pull:[observations] Fetched 400 of 11238 observations
    ```
    This PR introduces a fix by adding a `total_results` var.

- For huge media downloads, the script was stuck at this step forever and a half:
    ```
    INFO:f.connectors.inaturalist.inaturalist_pull:Starting media downloads: 18374 photo(s) and 46 sound(s) across 11438 observation(s).
    ```
    So this PR also adds a periodic logline about media download progress, every 200 attachments.

- Added a `max_month_lookback` param (similar to what is done for Global Forest Watch) to be able to limit queries to latest data only. And a little note about taking time to understand rate limiting.

## LLM use disclosure

None


Atmosfearful (approved):
lgtm, just curious though why we use "month lookback" and not a `start_timestamp` or start/end range which is more conventional

rudokemper:
Hm, I think it's an artifact of the very first time we implemented this, for `alerts_gcs`, which are published on a monthly cadence. And then we just kept that same convention. Fair to point out that there are better choices here. I might opt for `start_date` as that's more clear to non-technical folks. 

We don't do an end range because the connector scripts almost always are scheduled or triggered via webhook, so getting the latest data is always desirable.

rudokemper:
Actually. there is a better answer.  If we set a static start range, then the script(s) will always fetch from that date onward. And that can lead to a bunch of superfluous API calls for non-changing data. In the case of GFW, that data never changes once published. Not sure about iNaturalist; but in both cases, it's a _lot_ of data to have to fetch.

So setting a lookback window (coupled with a good cron schedule) is a way to ensure we only pull fresh data.

Interesting to think about this in relation to the different write strategies you had us thinking about today...

But I propose we punt on doing anything to modify this, until we get there with revisiting how connectors work.

Issue #196, opened 2026-01-08
Define deterministic primary key strategy for append uploads across connectors


rudokemper:
## Feature Request

Connector-based ingestion relies on primary keys to determine whether records should be inserted or updated. When source data provides a stable identifier, this works well. However, many raw datasets (e.g. GeoJSON uploads via the Dataset Importer) do not include feature IDs.

As of #136, the system generates **random UUIDs** for such records. This avoids baking in brittle assumptions about record equality and prevents subtle duplication bugs caused by small, hard-to-reason-about differences in feature content (e.g. null vs omitted fields, minor attribute edits). A TODO was intentionally left in the code to revisit this problem once append workflows are in scope.

However, supporting append-to-existing-dataset workflows in the future will require a more deliberate and explicit strategy for record identity. Naïve deterministic hashing of feature content is likely insufficient and may lead to surprising behavior or silent duplication. Equality semantics, scope of comparison, and user intent all need to be carefully defined before changing ID behavior at a low level.

This issue exists to scope and design a deterministic primary key strategy suitable for append workflows, without prematurely committing to a specific hashing or deduplication mechanism.

### User Stories & Acceptance Criteria

* **As a user appending data to an existing dataset**, I want record identity to behave predictably and not create silent duplicates due to minor, unintentional changes in source data.

Atmosfearful:
Still deep in the Discovery and learning phase but made decent progress. I'll consolidate today's notes as a checkpoint / notes-to-self to pick up on Monday.

---

First, I don't think there is any *primary key strategy* we can adopt that would on it's own allow us to confidently deduplicate/upsert/append. The issue is really just defining what "identity" means - it is too dangerous to make assumptions without any user guidance.

For example - our code sees "same geometry with different attributes", or "same attributes with different geometry"... deciding whether the user expects an update (deduplicated) or an append, is entirely context dependent.

- *With one Caveat, maybe it is possible to use the existing "source" hint that they provide to make some assumptions about specific tools; like "CoMapeo never allows duplicate geometry", or "Kobo always marks updates with some attribute XYZ", but in any case we would still need a solution for arbitrary csv upload, so I will ignore this avenue for now.*

So, this issue necessarily requires some changes to the upload UX so the user can tell us what "identity" means. I haven't decided which I like the most yet.
- They could identify one stable dedupe column/attribute, or several as a composite key
[...]

PR #300, opened 2026-09-16
Dataset Importer v2

Atmosfearful:
## Goal
Version 2 of the Dataset Importer app. This new solution accepts any source dataset (csv, geojson, etc.), and works for *any target dataset* (including connector-generated data).

It also enables the user to:

- Specify their import strategy (merge, append, sync, create)
- Define record identity (stable id, composite key, none)
- Review changes before committing them

**User Stories / JTBD**
*I want to manually import a GeoJSON that doesn't have ids, and merge it with an existing dataset*
Originally inferred from Rudo's description in #196 -> "many raw datasets (e.g. GeoJSON uploads via the Dataset Importer) do not include feature IDs"

V2 resolves this story by enabling the user to define the identity of their imported records.

## Screenshots
coming soon

## What I changed and why
todo

No database migration
Not touching shared python code (low regression risk)
Flexible solution (not aware of any unsupported upload user stories yet)
Backwards compatible with older datasets

## What I'm not doing here
No db migrations or major schema changes

Known limitations:
- I could have included a "replace dataset" option but decided against it just for simplicity. The difference between "sync" and "replace" is that "sync" retains existing columns even if they don't exist in the newly imported dataset. If you want to "replace" you can just delete the dataset and create a new one.
- `.geojson` imports with `GeometryCollections` are **not supported** because we don't have a valid table schema to support these. We would need `__geometry` instead of `__coordinates`.

## LLM use disclosure
Astra, Sol, Terra, Luna, the whole gang was involved ;D


.circleci/config.yml
```diff
@@ -6,8 +6,23 @@ workflows:
...
+      - image: cimg/node:22.22
+    steps:
+      - checkout
+      - run:
+          name: Install Dataset Importer dependencies
+          command: npm ci
+          working_directory: f/apps/gc_dataset_importer_v2.raw_app
+      - run:
+          name: Validate Dataset Importer app
+          command: npm run typecheck && npm run format:check && npm run lint && npm test && npm run lint:windmill && npm audit
```
Atmosfearful: Note for reviewers (and to future self)
perhaps we should generalize this job to "test-windmill-apps", but for now this is fine-- key point is we have a job separate from the existing gc scripts test run

f/apps/gc_dataset_importer_v2.raw_app/backend/apply_import.py
```diff
@@ -0,0 +1,5 @@
+from f.common_logic.dataset_importer_v2 import apply_import
+
+
+def main(db, import_id: str, preview_id: str):
```
Atmosfearful: This pattern is new to me; that these 'backend scripts' are individual jobs ran by windmill serverside. So that's why we define 5 separate ones (one for each step in the journey)

Each is just a thin wrapper around the dataset_importer_v2 module's functions

f/common_logic/db_operations.py
```diff
@@ -52,6 +52,37 @@ def conninfo(db: postgresql):
     return conn + password_part
 
 
+def column_mapping_table_name(table_name):
+    """Return the established mapping-table name for a dataset."""
+    return f"{table_name[:54]}__columns"
+
+
+def ensure_column_mapping_table(cursor, table_name, schema=None, upgrade_types=False):
```
Atmosfearful: legacy tables didn't have these mappings, so we lazily create them when using the v2 importer and targeting a legacy dataset

Why we need this: the importer can recognize the same original fields and use the same stored columns, even though their SQL-safe names would otherwise collide. It also matters for Merge/Sync identity: a selected source field can use the table to resolve to the correct column

PR #302, opened 2026-09-17
Add `application/geo+json` MIME type to dataset uploader

### LINKED ISSUE #295: Add `application/geo+json` MIME type to dataset uploader
## Current Buggy or Suspicious Behavior
.geojson files weren't appearing in my OS filebrowser because the extension of my file was .geojson not .json

#### Steps to Reproduce
Try to upload a .geojson file in step 2 of the dataset uploader app journey
Only files like CSV were visible by default (must change filter in OS level to find the .geojson)

## Expected Behavior
.geojson files should appear alongside .csv

rudokemper:
Closes https://github.com/ConservationMetrics/gc-scripts-hub/issues/295.

@Atmosfearful for future tiny fixes like this one, feel free to submit a PR directly too, it's probably just as much time to fix the problem as to write it up :slightly_smiling_face:


Atmosfearful:
We need some bots we can assign to stuff like this, so they can react immediately with a PR and I don't have to context switch!

> it's probably just as much time to fix the problem

Old habits / scar tissue... I used to grumble at people for changing things that never had a ticket. Occasionally the tiny fix turns out not to be so tiny

rudokemper:
As the de facto chief grumbler (aka PM for this work), I totally get you, and I probably _should_ have the same stance :sweat_smile:  I guess I'm trying to reduce friction to getting stuff done, but do hear you on the context switch too. Whatever works is good tbh.

Agree on bots!

PR #304, opened 2026-09-17
Add GBIF connector (windmill Flow)

### LINKED ISSUE #285: Add a GBIF connector to download occurrence data intersecting with a geometry
## Feature Request

Let's write a connector `f/connectors/gbif/gbif_pull.py` that takes a polygon as an input and uses the [Downloads API](https://techdocs.gbif.org/en/data-use/api-downloads) to download all GBIF occurrence data within that territory, allowing the community to explore and analyze the records directly in Guardian Connector.

So, it turns out that a GBIF download script will **not** be a run-of-the-mill connector script, as the GBIF Download API is an asynchronous job API (submit -> poll -> retrieve). A POST request triggers GBIF to asynchronously generate a download rather than immediately returning occurrence data. The initial request returns a download key, which then has to be polled until the job is complete. Once the status is `SUCCEEDED`, the resulting archive can be downloaded and processed. 

This means the script needs to behave more like a small workflow: submit the request, track its state, wait for completion, then ingest the finished dataset.

Is there a more native way to work with asynchronous job requests in Windmill, than writing code that polls for status every 30 seconds or something? Not sure! This has actually not come up before. 

### API information

Relevant endpoint: https://techdocs.gbif.org/en/data-use/api-downloads#requesting-an-occurrence-download

The predicate we want to use for geometry is [`within`](https://techdocs.gbif.org/en/data-use/api-downloads#within), which takes a WKT Polygon.

For a lookback window, we can use [`greaterThanOrEquals`](https://techdocs.gbif.org/en/data-use/api-downloads#greaterthanorequals) which can use the `LASTINTERPRETED` field that stores a timestamp; it is akin to "last modified".

#### Submit a request

```bash
curl --include \
  --user 'YOUR_GBIF_USERNAME:YOUR_GBIF_PASSWORD' \
  --header 'Content-Type: application/json' \
  --data '{
    "notificationAddresses": ["YOUR_EMAIL@example.com"],
    "sendNotification": true,
    "format": "SIMPLE_CSV",
    "predicate": {
      "type": "within",
      "geometry": "$BOUNDING_BOX"
        {
          \"type\": \"greaterThanOrEquals\",
          \"key\": \"LAST_INTERPRETED\",
          \"value\": \"$LOOKBACK_DATE\"
        }
    }
  }' \
  'https://api.gbif.org/v1/occurrence/download/request'
```

#### Check status

```bash
curl -s \
  'https://api.gbif.org/v1/occurrence/download/0001121-260916113435855' \
  | jq -r '.status'
```

For a geometry like `POLYGON((-55.30 3.10,-53.85 3.10,-53.85 3.80,-55.30 3.80,-55.30 3.10))` and no lookback value set, a request took about 16 minutes to compile a CSV of 53,586 records.

#### Once `SUCCEEDED`, download

```bash
curl -L \
  'https://api.gbif.org/v1/occurrence/download/request/0001121-260916113435855.zip' \
  -o gbif.zip
```

That will contain a giant CSV, which can be written to disk with `save_csv_to_postgres`.

### Possible script parameters

- `gbif_account` - We will need a [Windmill resource](https://www.windmill.dev/docs/core_concepts/resources_and_types) for username and password. For testing purposes, we can use a different resource that contains a pair like this? Or hell, just have them be strings at first, like what was done for Sensing Clues.
- `bounding_box` as a JSON list of coordinates `[[west, south], [east, north]]` which is a convention also used in the iNaturalist and Global Forest Watch connectors. I recommend adding a function to `f/common_logic/geoUtils` that converts that to WKT, which the GBIF script can then call.
- `max_months_lookback` - same conventions, descriptions, etc. as the iNaturalist and Global Forest Watch connectors. Use to construct a timestamp value for the `greaterThanOrEquals` predicate.
- `db_table_name` - same conventions, descriptions, etc. as all connector scripts
- `attachment_root` - same conventions, descriptions, etc. as all connector scripts

### Other tips

- Please request a real GBIF fixture for tests. I usually use `curl` to request something to play with, typically in my area around where I live as it's benign.
- Please check out other READMEs for what kind of stuff we typically put in there. We don't tend to go overboard with verbosity about how the script works, expecting the consumer of this repo to be able to read the code for themselves. I would document anything like API limitations, or script scheduling recommendations.
- Let's add a sane recommendation for a cron schedule to the README and YAML. Maybe once a month is good for this kind of dataset.
- Please check out other connector code for our conventions for Python docstrings, etc.
- Like the iNaturalist script, this one may end up taking longer than a Windmill instance job limit default (30m). So that might need to be documented as well.

### User Stories & Acceptance Criteria

From @abfleishman;

> I picture [that this occurrence data from GBIF can be helpful for the following reasons]:
> 1. mapping it.  with filters to see "where are the jaguar records" 
> 2. comparing it to recent data collection efforts by the partner
> 3. look at trends/patterns similar to what you can do on iNat 

See also https://docs.guardianconnector.net/guides/biodiversity/guide-gbif/#why-might-community-organizations-want-gbif-data

Additional context: https://github.com/ConservationMetrics/gc-roadmap/issues/19

Atmosfearful:
## Goal
Add an invokable or schedulable Windmill Flow that downloads and ingests species data from GBIF's data api.

The nature of the GBIF api requires this async approach, because it takes several minutes to generate even tiny datasets, and does not have any push/SSE/websocket behavior - must simply poll until the .zip is ready.

.zip bombs and OOM are avoided by limiting the map extent to 15,000 km2 (mainly just trying to mitigate a "select the entire planet" attack)

GBIF `gbifID` is copied to `_id`, so each run should safely "merge" correctly. Removed records won't be deleted (similar to all other connectors)

Staging link:
https://windmill.cmistaging.guardianconnector.net/flows/get/f/connectors/gbif/gbif_download

## Screenshots
image
test_gbif2

## What I changed and why
Added a Flow, which is a mix of YAML config and separate python scripts (one for each discrete step in the flow);

Includes a while-loop step that reaches one of these three end-cases;
1. Success: check returns `{"succeeded": true}`
2. Timeout: check compares the current time to the submission deadline, normally 24 hours after submission. It raises `TimeoutError`; with `skip_failures: false`, the Flow fails and does not ingest.
3. Terminal/error status: `CANCELLED`, `FAILED`, `KILLED`, `FILE_ERASED`, unknown status, or exhausted retryable HTTP failures raise an error.

Evaluated a few approaches including a simple state machine, Windmill State/Variables, and settled on Flow so that we can avoid adding a bespoke pg table.

See discussion in original issue ticket discussion thread #285 

Main failuremodes and retry behavior is covered, can't think of any other edge cases.

## What I'm not doing here
Not changing StructuredDBWriter or anything shared, so regression risk is low

## LLM use disclosure
Astra High for Plan and Review - Terra for impl. - via OpenCode TUI


f/connectors/gbif/README.md
```diff
@@ -0,0 +1,35 @@
+# `gbif_download`: GBIF occurrences
```
Atmosfearful: Note to self to clean this up and make it more human, I'm not a fan of these super dense LLM readmes

f/connectors/gbif/gbif_download.flow/flow.yaml
```diff
@@ -0,0 +1,133 @@
+summary: 'GBIF: Download occurrences'
+description: 'Submits one GBIF occurrence download, polls it with Windmill
```
Atmosfearful: note to self to make this more human readable

f/connectors/gbif/gbif_submit_download.script.yaml
```diff
@@ -0,0 +1,39 @@
...
+      format: resource-c_gbif
+    max_months_lookback:
+      type: integer
+      description: Optional non-negative months of LAST_INTERPRETED history, from the
+        first day of the cutoff month.
+      default: null
+    max_wait_seconds:
+      type: integer
+      description: Maximum elapsed seconds allowed for Flow polling, from 1 through 86400.
+      default: 86400
```
rudokemper: I wonder about the use case in which somebody might want to change this. Maybe we just hard code 86400 seconds (24 hours)?

If we **do** keep it, then I think we should ask for hours, or at least minutes, and do a line of math to turn it into seconds.

f/connectors/gbif/gbif_submit_download.script.yaml
```diff
@@ -0,0 +1,39 @@
...
+        Location Helper: https://labs.mapbox.com/location-helper
+      default: null
+    gbif_account:
+      type: object
+      description: GBIF account resource containing its username (not email) and password.
+      default: null
+      format: resource-c_gbif
+    max_months_lookback:
+      type: integer
+      description: Optional non-negative months of LAST_INTERPRETED history, from the
```
rudokemper: Give a short description of what LAST_INTERPRETED means according to GBIF.

(I am commenting on the scripts but the same changes should be made to the Flow, which is what the user will see.)

f/connectors/gbif/gbif_submit_download.script.yaml
```diff
@@ -0,0 +1,39 @@
...
+      description: >-
+        Two corners as `[[west, south], [east, north]]` in longitude/latitude
+        order; limited to approximately 15,000 km2. Copy bounds from Mapbox
+        Location Helper: https://labs.mapbox.com/location-helper
+      default: null
+    gbif_account:
+      type: object
+      description: GBIF account resource containing its username (not email) and password.
+      default: null
+      format: resource-c_gbif
```
rudokemper: Note: our workflow has been to submit a resource to the Windmill Hub so that these can be auto-synced to each instance. Else, we have to create the resource X amount of times for every instance.

https://hub.windmill.dev/resource_types/add

I find that Windmill team then needs a ping on Discord to approve it. They have yet to not approve one of my submissions.

(Fine to keep `c_gbif` here, and update with the official resource type once approved.)

f/connectors/gbif/gbif_submit_download.script.yaml
```diff
@@ -0,0 +1,39 @@
+summary: 'GBIF: Submit occurrence download'
+description: Validates a small longitude/latitude bounding box and submits one
```
rudokemper: For each of the script descriptions, I might add a note to say that the scripts typically are not run on their own ever. And to open the flow instead.

f/connectors/gbif/gbif_pull.script.yaml
```diff
@@ -0,0 +1,40 @@
+summary: 'GBIF: Download and import occurrences'
+description: Downloads a completed GBIF SIMPLE_CSV archive, preserves
```
rudokemper: I think at some point we should do a language "style guide" for writing user-facing descriptions. While this is totally correct, I doubt a single one of our users knows what "upserting" is.

(This is a repo-wide problem so not something to change here, I'm sure most connectors suffer from the same challenge.)

f/connectors/gbif/gbif_pull.py
```diff
@@ -0,0 +1,184 @@
...
+    try:
+        longitude_value = float(longitude)  # type: ignore[arg-type]
+        latitude_value = float(latitude)  # type: ignore[arg-type]
+    except (TypeError, ValueError):
+        return None
+    if not (math.isfinite(longitude_value) and math.isfinite(latitude_value)):
+        return None
+    if not -180 <= longitude_value <= 180 or not -90 <= latitude_value <= 90:
+        return None
+    return [longitude_value, latitude_value]
```
rudokemper: I might do a quick check if we have a similar validation elsewhere, since so much of our code is about spatial data.

Me personally, I might even add it to `geoUtils` so it's there to be found / discovered for future coord validations.

f/connectors/gbif/gbif_pull.py
```diff
@@ -0,0 +1,184 @@
...
+        with zipfile.ZipFile(archive) as zipped:
+            members = [
+                member
+                for member in zipped.infolist()
+                if member.filename.lower().endswith(".csv") and not member.is_dir()
+            ]
+            if len(members) != 1:
+                raise ValueError(
+                    "GBIF archive must contain exactly one occurrence CSV member."
+                )
```
rudokemper: I think for other connectors where the total records = 0, we don't raise an error / fail the job, we just skip writes and return nothing. That gives the user a sense of "okay, things were configured correctly."

Imagine a use case where somebody is about to do an expedition in the middle of nowhere where no data has been collected, and they want to schedule this connector in advance. We take a similar approach to alerts.

Now, if GBIF itself refuses to produce an archive with 0 occurrences, then that's different as we won't have a zipfile to begin with.

f/connectors/gbif/gbif_pull.py
```diff
@@ -0,0 +1,184 @@
...
+    provenance = {
+        "download_key": download_key,
+        "doi": metadata.get("doi"),
+        "license": metadata.get("license"),
+        "predicate": request.get("predicate") if isinstance(request, dict) else None,
+        "record_count": record_count,
+    }
+    (destination / f"{download_key}.json").write_text(
+        json.dumps(provenance, indent=2), encoding="utf-8"
+    )
```
rudokemper: love this.

f/connectors/gbif/gbif_pull.py
```diff
@@ -0,0 +1,184 @@
...
+    """
+    if (
+        not db_table_name
+        or len(db_table_name) > 54
+        or "/" in db_table_name
+        or "\\" in db_table_name
+    ):
+        raise ValueError(
+            "db_table_name must be a non-empty table name of at most 54 characters."
+        )
```
rudokemper: Hm, we don't do this check anywhere else. But maybe we should?

I'd propose to file an issue to introduce a util fn for `db_table_name` validations.

f/connectors/gbif/gbif_check_download.py
```diff
@@ -0,0 +1,56 @@
...
+
+"""Perform one GBIF download status check for a Flow polling loop."""
+
+from datetime import datetime, timezone
+
+import requests
+
+_API = "https://api.gbif.org/v1/occurrence/download"
+_PENDING = frozenset({"PREPARING", "RUNNING", "SUSPENDED"})
+_TERMINAL = frozenset({"CANCELLED", "KILLED", "FAILED", "FILE_ERASED"})
```
rudokemper: If available, a quick comment with a link to the relevant docs page where these are defined is always good.

f/connectors/gbif/gbif_download.flow/flow.yaml
```diff
@@ -0,0 +1,133 @@
...
+              path: f/connectors/gbif/gbif_check_download
+            retry:
+              exponential:
+                attempts: 3
+                multiplier: 2
+                random_factor: 20
+                seconds: 5
+              retry_if:
+                expr: error.message && error.message.startsWith('retryable GBIF status check
+                  failure')
```
rudokemper: Oh great, this is what I was hoping for!

f/common_logic/geo_utils.py
```diff
@@ -1,12 +1,82 @@
...
 import logging
+import math
 import tempfile
 from pathlib import Path
 
 logging.basicConfig(level=logging.INFO)
 logger = logging.getLogger(__name__)
 
 
+def bounding_box_to_wkt(
```
rudokemper: Consider using `box` from `shapely` and `Geod` from `pyproj` to simplify this code.

(We already use both across this repo.)

f/connectors/gbif/gbif_pull.py
rudokemper: This would be a great unit test (to add)

f/connectors/gbif/README.md
```diff
@@ -0,0 +1,35 @@
...
+
+Run the `f/connectors/gbif/gbif_download` Flow for an initial backfill or an
+ad-hoc update. It submits a GBIF asynchronous SIMPLE_CSV download, waits
+passively for completion, preserves the archive and citation metadata, then
+upserts its occurrences into PostgreSQL.
+
+Create a Windmill `c_gbif` resource manually (custom resource types cannot be
+synced): `{"username":"string","password":"string"}`. Use the GBIF
+**username**, not an email address. Never place credentials in Flow inputs or
+logs. Register an account and review [GBIF download API restrictions](https://techdocs.gbif.org/en/data-use/api-downloads), including its load-dependent limit on incomplete downloads per account.
```
rudokemper: Add a TODO to create the resource on Windmill Hub

f/connectors/gbif/README.md
```diff
@@ -0,0 +1,35 @@
...
+
+Bounds must be `[[west, south], [east, north]]` in longitude/latitude order,
```
rudokemper: Add a > [!TIP] linking to the Mapbox location helper. I think I did this for iNaturalist too.

f/connectors/gbif/README.md
rudokemper: It's better than the verbose slop I'm used to from LLMs! But yeah - I recommend taking a look at a few other READMEs to see some styles and conventions.

rudokemper (changes requested):
Generally really great. I left comments throughout, but they are all pretty minor fixes that I think can be easily resolved in one round of revisions.

My biggest concern that I did not leave in comments is whether the hard-coded limit of 15,000 km2 is going to work out.

The user that inspired this script wanted to run it over their territory, which is almost exactly 25,000 km2. And because it's a remote area, via the [gbif map search](https://conservationmetrics.github.io/gbif-map-search) tool I see there are only 93,222 occurrences, so it won't be a ZIP bomb.

But we should have _some_ limit. And I doubt anyone will ever go higher than what this user wants to. Maybe we double the limit to 30,000 km2? (And if someone complains, we deal about it then. Or propose a workaround to schedule the Flow twice with different bboxes - maybe we can even have that as a > [!TIP] in the README where we describe the limit...

f/connectors/gbif/gbif_pull.py
```diff
@@ -0,0 +1,184 @@
...
+def _stream_download(url: str, destination: Path) -> None:
+    with requests.get(url, stream=True, timeout=(10, 300)) as response:
+        response.raise_for_status()
+        with destination.open("wb") as output:
+            for chunk in response.iter_content(chunk_size=1024 * 1024):
+                if chunk:
+                    output.write(chunk)
+
+
+def _convert_archive(archive: Path, output_path: Path) -> tuple[Path, int]:
```
rudokemper: Call `camel_to_snake` from `identifier_utils.py` to convert camelCase to snake_case? We write snake case in every other connector.

rudokemper:
Oh, one additional thing I am noticing:

A few columns provide a UUID value:
image

image

These are actually really important pieces of information. As one of the first questions users want to know are 
- "what are the biggest datasets of biodiversity data in my territory?"
- "who are the publishers of data holding biodiversity data?"

I expect we can use the same Occurrence API to resolve these: https://github.com/conservationmetrics/gbif-map-search

Then we can add a column `dataset` and `publishing_org` and maybe remove the uuid key fields.

Totally fine for this to be a separate issue, and PR.

f/connectors/gbif/gbif_submit_download.script.yaml
Atmosfearful: Good question, I left it configurable as I was unsure what the longest download might be. But in hindsight that is silly because it just pushes that question to the user. So a better idea is to hard code it and wait for a failure report before changing it

f/connectors/gbif/gbif_submit_download.script.yaml
Atmosfearful: great, created a ticket #305

f/connectors/gbif/gbif_pull.script.yaml
Atmosfearful: Ah yeah even for me, I forget that less-technical people will invoke these.

Possibly a good fit for AGENTS.md - for now I'll add it to my devcontainer's agent rails, which I plan to share eventually once it's stable and battle tested. I am still slowly forming opinions about stuff like, this as I get oriented (docs, styleguides, agent rails, skills, mcp).

f/connectors/gbif/gbif_pull.py
Atmosfearful: This code explicitly handles an error case where the API gave us no csv at all-- I'll improve the error message to make that clear. The current phrasing of 'occurence CSV member' is misleading.

In my empty ocean test with 0 occurrences, we still get a csv but it has zero rows. And that is handled silently as noop, which I believe is what we want. Logs capture that as `"record_count": 0`

Should already have test coverage but ill double check.

Sample run:
https://windmill.cmistaging.guardianconnector.net/run/01a0c5f9-edab-872a-1aab-d4c62b32c362?workspace=cmistaging

f/connectors/gbif/gbif_pull.py
Atmosfearful: good catch, also adding as a note to my agent guardrails

f/connectors/gbif/gbif_pull.py
Atmosfearful: https://github.com/ConservationMetrics/gc-scripts-hub/issues/306

f/connectors/gbif/README.md
Atmosfearful: I created a ticket - https://github.com/ConservationMetrics/gc-scripts-hub/issues/305
I will update the readme when the official resource is created

Separate discussion but I will propose that we ban inline TODOs anywhere in any github repos :policeman:

f/connectors/gbif/gbif_pull.py
Atmosfearful: good catch, also adding as a note to my agent guardrails

PR #309, opened 2026-09-22
GBIF: resolve dataset names and publishing org names

### LINKED ISSUE #307: [GBIF] Resolve Dataset and Publishing Org Keys
From Rudo:

Oh, one additional thing I am noticing:

A few columns provide a UUID value (DATASET KEY and PUBLKISHING_ORG_KEY)

These are actually really important pieces of information. As one of the first questions users want to know are 
- "what are the biggest datasets of biodiversity data in my territory?"
- "who are the publishers of data holding biodiversity data?"

I expect we can use the same Occurrence API to resolve these: https://github.com/conservationmetrics/gbif-map-search

Then we can add a column `dataset` and `publishing_org` and maybe remove the uuid key fields.

Totally fine for this to be a separate issue, and PR.

_Originally posted by @rudokemper in https://github.com/ConservationMetrics/gc-scripts-hub/issues/304#issuecomment-5734396364_

Atmosfearful:
## Goal
Resolve UUIDs to names using the GBIF API
- UUID columns `dataset_key` and `publishing_org_key` are preserved
- Adds new columns, `dataset` and `publishing_org` which are nullable strings
- How it works: deduplicates the UUIDs and batches requests to the API to resolve them. Stores as a separate "enriched" csv in the datalake. The final enriched csv is imported into the warehouse db.
- API outages / network issues fail by falling back to default behavior. The import still completed but with unresolved dataset / publishing_org names.

Known limitations:
GBIF does not publish a rate limit policy but say that they may return 429 errors if their servers are under heavy load. In this case we just treat it like other failures and write `NULL` to the resolved name columns. The user can try importing again at a later date. Resolving this would add complex retry logic that I felt was out of scope.

Closes #307 

## Screenshots
Screenshot_2026-09-22_11-54-55
Screenshot_2026-09-22_11-55-03

## What I changed and why
Added logic and tests to existing importer flow

changes stay confined to the import flow and tests, and fail gracefully back to the default behavior (no uuid resolution)
QAd by doing an import to an existing dataset and rendering in the map explorer

## What I'm not doing here
See above limitations

## LLM use disclosure
Researched / Planned / Reviewed with Astra High generated with Terra Default via OpenCode TUI


rudokemper (approved):
Great - thank you!

